PR #8, opened 2024-10-31
Adds bundle support to block builder

dylanlott:
This PR adds a `BundleFetcher` that the block builder uses to fetch bundles from the tx cache. It works similarly to the tx poller:
- it requests all bundles from the cache
- it filters out the already seen ones, keeping only the unique bundles
- it then sends these to the block task, which ingests the bundles one by one.
  - When ingesting the bundles, right now only the txs are ingested, without respecting the set configurations for bundles, nor including the permit2 signed orders. This means that only the basic bundle functionality is implemented. Further implementation probably requires simulation work.
  - There's no specific rule to where the txs of the bundle land in the block. We are simply extending the block with the new transactions.

Closes ENG-630.

DO NOT MERGE until ENG-633 is implemented.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,7 +44,7 @@ openssl = { version = "0.10", features = ["vendored"] }
 reqwest = { version = "0.11.24", features = ["blocking", "json"] }
 ruint = "1.12.1"
 serde_json = "1.0"
-thiserror = "1.0.58"
+thiserror = "1.0.68"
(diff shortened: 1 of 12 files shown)


Evalir:
Updated description and broke out tasks into separate tickets—this now implements the basic functionality to make bundle fetching / block building with it work.

Evalir (approved):

PR #19, opened 2024-11-08
feat(`oauth`): caching, shareable authenticator

Evalir:
Implements a shareable `Authenticator` which periodically refreshes the token it holds through a spawnable task. The interval for refreshing the token is configurable.

We share this authenticator across all the builder tasks that need it, and spawn the refresh task along with all other handles.

Closes ENG-633


dylanlott (approved):
:shipit:

PR #22, opened 2024-12-05
refactor: deprecate tx & bundle polling tasks

anna-carroll:


src/tasks/block.rs
```diff
@@ -109,38 +110,69 @@ impl InProgressBlock {
...
-        let handle = tokio::spawn(
+        tokio::spawn(
             async move {
                 loop {
 
                     select! {
                         biased;
                         _ = &mut sleep => {
+                            // Build a block
+                            let mut in_progress = InProgressBlock::default();
```
dylanlott, line 172: Nice, easier to reason about one loop than three for sure 🎀

src/tasks/block.rs
```diff
@@ -109,38 +110,69 @@ impl InProgressBlock {
...
+    async fn get_transactions(&mut self, in_progress: &mut InProgressBlock) {
+        let txns = self.tx_poller.check_tx_cache().await;
+        match txns {
+            Ok(txns) => {
+                for txn in txns.into_iter() {
+                    in_progress.ingest_tx(&txn);
+                }
+            }
+            Err(e) => {
+                tracing::error!("error polling transactions: {:?}", e);
```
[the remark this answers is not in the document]
dylanlott: Nit: avoid string interpolation

src/tasks/block.rs
```diff
@@ -109,38 +110,69 @@ impl InProgressBlock {
...
+    async fn get_bundles(&mut self, in_progress: &mut InProgressBlock) {
+        let bundles = self.bundle_poller.check_bundle_cache().await;
+        match bundles {
+            Ok(bundles) => {
+                for bundle in bundles {
+                    in_progress.ingest_bundle(bundle);
+                }
+            }
+            Err(e) => {
+                tracing::error!("error polling bundles: {:?}", e);
```
dylanlott, line 152: Same as above, avoid string interpolation

src/tasks/tx_poller.rs
```diff
@@ -81,37 +79,4 @@ impl TxPoller {
             self.seen_txns.remove(&key);
         }
     }
-
-    /// spawns a task that polls the tx-pool for unique transactions and ingests them into the tx_channel.
-    pub fn spawn(mut self, tx_channel: mpsc::UnboundedSender<TxEnvelope>) -> JoinHandle<()> {
```
dylanlott, line 86: The best asynchronous code is deleted asynchronous code 💣

dylanlott (approved):
No showstoppers, approving to unblock 🔥

src/tasks/block.rs
```diff
@@ -109,38 +110,69 @@ impl InProgressBlock {
...
 
                     select! {
                         biased;
                         _ = &mut sleep => {
+                            // Build a block
+                            let mut in_progress = InProgressBlock::default();
+                            self.get_transactions(&mut in_progress).await;
+                            self.get_bundles(&mut in_progress).await;
+
                             if !in_progress.is_empty() {
```
Evalir, line 176: So, as far as I understand now, instead of spawning a loop that awaits 3 different futures concurrently, we now have a loop that:
1. gets all txs
2. gets all bundles
3. builds the block
4. sends it
5. sleeps for X seconds

In this case we don't need the `select!` at all anymore. We just need to loop like we already are and go through the steps.

src/tasks/block.rs
anna-carroll: https://github.com/init4tech/builder/pull/22/commits/72f389c971c345b32fe2bec267d866a806b161c6

src/tasks/block.rs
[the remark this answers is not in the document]
anna-carroll: https://github.com/init4tech/builder/pull/22/commits/72f389c971c345b32fe2bec267d866a806b161c6

Evalir (changes requested):

src/tasks/block.rs
anna-carroll: good shout https://github.com/init4tech/builder/pull/22/commits/ada0cd28cb6ab1585cb7e978809448f7a6980b07

src/tasks/block.rs
```diff
@@ -109,75 +110,78 @@ impl InProgressBlock {
...
+                    tokio::time::sleep(Duration::from_secs(self.incoming_transactions_buffer))
+                        .await;
+
+                    // Build a block
+                    let mut in_progress = InProgressBlock::default();
+                    self.get_transactions(&mut in_progress).await;
+                    self.get_bundles(&mut in_progress).await;
+
+                    // submit the block if it has transactions
+                    if !in_progress.is_empty() {
```
Evalir, line 174: should we maybe log if we find there are no txs?

src/tasks/block.rs
```diff
@@ -109,75 +110,78 @@ impl InProgressBlock {
...
-                                }
-                            }
+                    // sleep the buffer time
+                    tokio::time::sleep(Duration::from_secs(self.incoming_transactions_buffer))
+                        .await;
+
+                    // Build a block
+                    let mut in_progress = InProgressBlock::default();
+                    self.get_transactions(&mut in_progress).await;
+                    self.get_bundles(&mut in_progress).await;
```
Evalir, line 171: an interesting byproduct of this is that bundles will land end of block. nothing to do here now, but something we might wanna think about later

Evalir (approved):
cool!

src/tasks/block.rs
anna-carroll: i have this in next PR :)

src/tasks/block.rs
anna-carroll: yes i think any logic around ordering could be layered in later

PR #23, opened 2024-12-06
fix: update builder loop to wake up at target block time

anna-carroll:
this PR makes the sleep time “block-slot-aware”

it targets a configurable interval within the block slot at which it begins building a new block

previously, sleeping a fixed buffer amount will cause the builder to attempt to submit blocks at random times relative to the chain’s block slot time; e.g. it could be attempting to submit 1sec before the slot ends, in which case the submission will almost certainly fail


Evalir (approved):
solid. just need to update the new env vars, no?

PR #29, opened 2024-12-09
fix: remove bundle polling for now; update logging level

anna-carroll:


src/tasks/block.rs
```diff
@@ -177,23 +181,25 @@ impl BlockBuilder {
...
                     tokio::time::sleep(Duration::from_secs(self.secs_to_next_target())).await;
-                    tracing::trace!("beginning block build cycle");
+                    tracing::info!("beginning block build cycle");
 
                     // Build a block
                     let mut in_progress = InProgressBlock::default();
                     self.get_transactions(&mut in_progress).await;
-                    self.get_bundles(&mut in_progress).await;
+
+                    // TODO: Implement bundle ingestion #later
```
Evalir, line 190: let's add a linear ticket for this

Evalir (approved):
lgtm

src/tasks/block.rs
anna-carroll: this work should be tracked somewhere i think? cc @dylanlott

PR #31, opened 2024-12-11
feat: filter confirmed transactions from the block

anna-carroll:
ensures that transactions which have already confirmed on the rollup are not included in the block

---
does not include removal of the "seen_txns" cache in the TxPoller class. going to make a separate PR for that

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -195,7 +195,7 @@ impl BuilderConfig {
         }
     }
 
-    /// Connect to rollup rpc provider.
+    /// Connect to the Rollup rpc provider.
     pub async fn connect_ru_provider(&self) -> Result<WalletlessProvider, ConfigError> {
         ProviderBuilder::new()
             .with_recommended_fillers()
@@ -204,7 +204,7 @@ impl BuilderConfig {
             .map_err(Into::into)
     }
 
-    /// Connect to an rpc provider.
+    /// Connect to the Host rpc provider.
     pub async fn connect_host_provider(&self) -> Result<Provider, ConfigError> {
         let builder_signer = self.connect_builder_signer().await?;
         ProviderBuilder::new()
--- a/src/config.rs
+++ b/src/config.rs
@@ -14,6 +14,7 @@ use zenith_types::Zenith;
 const HOST_CHAIN_ID: &str = "HOST_CHAIN_ID";
 const RU_CHAIN_ID: &str = "RU_CHAIN_ID";
 const HOST_RPC_URL: &str = "HOST_RPC_URL";
+const ROLLUP_RPC_URL: &str = "ROLLUP_RPC_URL";
 const TX_BROADCAST_URLS: &str = "TX_BROADCAST_URLS";
 const ZENITH_ADDRESS: &str = "ZENITH_ADDRESS";
 const QUINCEY_URL: &str = "QUINCEY_URL";
@@ -44,6 +45,8 @@ pub struct BuilderConfig {
     pub ru_chain_id: u64,
     /// URL for Host RPC node.
     pub host_rpc_url: Cow<'static, str>,
+    /// URL for the Rollup RPC node.
+    pub ru_rpc_url: Cow<'static, str>,
     /// Additional RPC URLs to which to broadcast transactions.
     pub tx_broadcast_urls: Vec<Cow<'static, str>>,
     /// address of the Zenith contract on Host.
@@ -139,6 +142,7 @@ impl BuilderConfig {
             host_chain_id: load_u64(HOST_CHAIN_ID)?,
             ru_chain_id: load_u64(RU_CHAIN_ID)?,
             host_rpc_url: load_url(HOST_RPC_URL)?,
+            ru_rpc_url: load_url(ROLLUP_RPC_URL)?,
             tx_broadcast_urls: env::var(TX_BROADCAST_URLS)
                 .unwrap_or_default()
                 .split(',')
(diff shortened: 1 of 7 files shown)


src/tasks/block.rs
```diff
@@ -161,6 +173,27 @@ impl BlockBuilder {
         self.bundle_poller.evict();
     }
 
+    async fn filter_transactions(&mut self, in_progress: &mut InProgressBlock) {
```
prestwich, line 176: does self need to be `&mut` here? can it not be `&self` ?

src/tasks/block.rs
```diff
@@ -62,6 +61,13 @@ impl InProgressBlock {
         self.transactions.push(tx.clone());
     }
 
+    /// Remove a transaction from the in-progress block.
+    pub fn remove_tx(&mut self, tx: &TxEnvelope) {
+        tracing::info!(hash = %tx.tx_hash(), "removing tx");
```
prestwich, line 66: a lot of these `info` seem too verbose. this seems like a `trace`

src/config.rs
```diff
@@ -180,13 +184,27 @@ impl BuilderConfig {
...
-    /// Connect to the provider using the configuration.
-    pub async fn connect_provider(&self) -> Result<Provider, ConfigError> {
+    /// Connect to host rpc provider.
+    pub async fn connect_host_provider(&self) -> Result<Provider, ConfigError> {
         let builder_signer = self.connect_builder_signer().await?;
+        BuilderConfig::connect_provider(builder_signer, self.host_rpc_url.clone()).await
+    }
+
+    /// Connect to rollup rpc provider.
+    pub async fn connect_ru_provider(&self) -> Result<Provider, ConfigError> {
```
prestwich, line 194: this is probably outside of scope, but given the builder never needs to submit rollup txns, it technically doesn't need a signer

src/tasks/block.rs
```diff
@@ -113,15 +119,21 @@ impl InProgressBlock {
 /// BlockBuilder is a task that periodically builds a block then sends it for signing and submission.
 pub struct BlockBuilder {
     pub config: BuilderConfig,
+    pub ru_provider: crate::config::Provider,
```
prestwich, line 122: can replace these fully-qualified paths with an import?

src/tasks/block.rs
```diff
@@ -113,15 +119,21 @@ impl InProgressBlock {
...
     pub bundle_poller: BundlePoller,
 }
 
 impl BlockBuilder {
     // create a new block builder with the given config.
-    pub fn new(config: &BuilderConfig, authenticator: Authenticator) -> Self {
+    pub fn new(
+        config: &BuilderConfig,
+        authenticator: Authenticator,
+        ru_provider: crate::config::Provider,
```
Evalir, line 132: ditto on imports

src/tasks/block.rs
```diff
@@ -161,6 +173,27 @@ impl BlockBuilder {
...
+            let tx = self
+                .ru_provider
+                .get_transaction_by_hash(*transaction.tx_hash())
+                .await
+                .expect("failed to get receipt");
+            if tx.is_some() {
+                confirmed_transactions.push(transaction.clone());
+            }
+        }
+        tracing::info!(confirmed = confirmed_transactions.len(), "found confirmed transactions");
```
Evalir, line 189: this should be a trace too, imo

src/tasks/block.rs
anna-carroll: gotcha, i was just going off of the same log level in `ingest_tx`. i can downgrade this to `trace`?

honestly in general i've questioned my instincts around what log levels to use.. any higher level guidance people have on this?

src/tasks/block.rs
anna-carroll: the compiler was complaining because we import 
```
use alloy::providers::Provider;
```
so it was necessary to disambiguate

however i don't see in the code where that import is being used.. let me try to remove it and swap it

src/tasks/block.rs
anna-carroll: https://github.com/init4tech/builder/pull/31/commits/349c96e4767781d50c0c7ca48ac26b28b06311bf

src/tasks/block.rs
anna-carroll: https://github.com/init4tech/builder/pull/31/commits/7371efa93700c4855e1505bf6f7c976ebf627baf

src/tasks/block.rs
anna-carroll: https://github.com/init4tech/builder/pull/31/commits/b3d3d982707a5b87ec163b7e3a8e3d9c6dc0baf9

src/tasks/block.rs
anna-carroll: https://github.com/init4tech/builder/pull/31/commits/b3d3d982707a5b87ec163b7e3a8e3d9c6dc0baf9

src/config.rs
anna-carroll: 🙂 https://github.com/init4tech/builder/pull/31/commits/daeb424e03d11440fe0c7b3985554b61fcc2da04

tests/bundle_poller_test.rs
```diff
@@ -20,7 +20,8 @@ mod tests {
         let config = BuilderConfig {
             host_chain_id: 17000,
             ru_chain_id: 17001,
-            host_rpc_url: "http://rpc.holesky.signet.sh".into(),
+            host_rpc_url: "host-rpc.example.com".into(),
+            ru_rpc_url: "ru-rpc.example.com".into(),
```
dylanlott, line 24: These RPC URLs need to be set to proper values, they'll fail with these example ones. They're not being run right now in the typical test suite, though, which is why they're passing CI.

tests/tx_poller_test.rs
```diff
@@ -67,7 +67,8 @@ mod tests {
         let config = BuilderConfig {
             host_chain_id: 17000,
             ru_chain_id: 17001,
-            host_rpc_url: "http://rpc.holesky.signet.sh".into(),
+            host_rpc_url: "host-rpc.example.com".into(),
+            ru_rpc_url: "ru-rpc.example.com".into(),
```
dylanlott, line 71: Ditto here.

tests/bundle_poller_test.rs
dylanlott: Discarding my own comment - these values aren't used anymore like I thought they were. Will approve to unblock now.

dylanlott (approved):

prestwich (approved):

PR #32, opened 2024-12-11
fix: remove naive seen_txns cache

anna-carroll:
Given #31 , it's now possible to remove the naive solution of the `seen_txns` cache, and only filter transactions by whether they've been included on the rollup or not.

### DIFF
--- a/src/tasks/block.rs
+++ b/src/tasks/block.rs
@@ -153,7 +153,6 @@ impl BlockBuilder {
                 tracing::error!(error = %e, "error polling transactions");
             }
         }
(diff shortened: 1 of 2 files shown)


anna-carroll:
As we can see, at this point the `TxPoller` struct is an extremely simple wrapper for one function that calls the tx cache endpoint. It might make sense to refactor this, to combine it with the `BundlePoller` struct into a single `TransactionCache` struct which has two methods: `get_transactions` and `get_bundles`

prestwich (approved):
straightforward

PR #36, opened 2024-12-20
feat: integration test

anna-carroll:
on a loop
- sends a transaction to the target chain
- waits for it to mine
- sleeps

while collecting metrics on mine time

---
env vars
```
RPC_URL=target chain RPC URL
CHAIN_ID=target chain ID
AWS_KMS_KEY_ID=any KMS key funded on the target chain 
RECIPIENT_ADDRESS=the address of the recipient of the Ether
SLEEP_TIME=60 (or whatever)
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,10 @@ name = "builder"
 name = "zenith-builder-example"
 path = "bin/builder.rs"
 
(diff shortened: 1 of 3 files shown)


dylanlott (approved):
LGTM

PR #34, opened 2024-12-12
[spike] Adds bundle helper submission call

dylanlott:
# [WIP] 

### TL;DR

This PR modifies the submit task in the Builder to call the bundle helper contract's submit function. 

### What changed?

- Introduced `FillPermit2` vector for transaction fills
- Replaced `Zenith::BlockHeader` with `BlockHeader` type
- Added `submitCall` structure with fills, header, and signature components
- Modified transaction building to use the bundle helper `submit` function
- Added error handling for transaction submission

PR #43, opened 2025-01-10
chore: bumps alloy and zenith versions

dylanlott:
- bump alloy up
- remove alloy primitives and import primitives from alloy core instead
- fix errors from alloy updates
- remove signet-types and reference zenith-types instead

PR #45, opened 2025-01-10
feat: adds bundles to in progress block creation

dylanlott:
# tl;dr 

This PR adds bundle ingestion from the tx-pool to the Signet Builder.

## Changes 

- `InProgressBlock` building process now checks the cache for bundles
- Ingests any new bundles that it finds into the in progress block
- Changes the block submission call to use the BuilderHelper contract
- Stubs out bundle simulation (see [ENG-640](https://linear.app/initiates/issue/ENG-640/add-simulation-engine-to-the-example-builder))

## Configuration updates

This PR requires an environment variable (the builder helper contract address) be added to the configuration, and thus must go in after the config is added to the environment in [the-infra PR #308](https://github.com/init4tech/the-infra/pull/308)

Closes [ENG-549](https://linear.app/initiates/issue/ENG-629/add-bundle-with-signed-orders-support-by-using-the-helper-contract-to)
Supersedes builder [PR #34](https://github.com/init4tech/builder/pull/34)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ path = "bin/submit_transaction.rs"
 [dependencies]
 zenith-types = "0.13"
 
-alloy = { version = "0.7.3", features = ["full", "json-rpc", "signer-aws", "rpc-types-mev"] }
+alloy = { version = "0.7.3", features = ["full", "json-rpc", "signer-aws", "rpc-types-mev", "rlp"] }
 alloy-rlp = { version = "0.3.4" }
 
 aws-config = "1.1.7"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,3 +51,6 @@ async-trait = "0.1.80"
 oauth2 = "4.4.2"
 metrics = "0.24.1"
 metrics-exporter-prometheus = "0.16.0"
+
+[dev-dependencies]
+httpmock = "0.7.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,6 +51,3 @@ async-trait = "0.1.80"
 oauth2 = "4.4.2"
 metrics = "0.24.1"
 metrics-exporter-prometheus = "0.16.0"
-
-[dev-dependencies]
-httpmock = "0.7.0"
--- a/bin/builder.rs
+++ b/bin/builder.rs
@@ -34,7 +34,7 @@ async fn main() -> eyre::Result<()> {
 
     let submit = SubmitTask {
         authenticator: authenticator.clone(),
-        host_provider,
+        host_provider: host_provider.clone(),
         zenith,
         client: reqwest::Client::new(),
         sequencer_signer,
@@ -44,7 +44,7 @@ async fn main() -> eyre::Result<()> {
 
     let authenticator_jh = authenticator.spawn();
     let (submit_channel, submit_jh) = submit.spawn();
-    let build_jh = builder.spawn(submit_channel);
+    let build_jh = builder.spawn(submit_channel, host_provider.clone());
 
     let port = config.builder_port;
     let server = serve_builder_with_span(([0, 0, 0, 0], port), span);
--- a/bin/builder.rs
+++ b/bin/builder.rs
@@ -27,14 +27,13 @@ async fn main() -> eyre::Result<()> {
     let sequencer_signer = config.connect_sequencer_signer().await?;
     let zenith = config.connect_zenith(host_provider.clone());
 
-    let builder = BlockBuilder::new(&config, authenticator.clone(), ru_provider);
-
     let metrics = MetricsTask { host_provider: host_provider.clone() };
     let (tx_channel, metrics_jh) = metrics.spawn();
 
+    let builder = BlockBuilder::new(&config, authenticator.clone(), ru_provider.clone());
     let submit = SubmitTask {
         authenticator: authenticator.clone(),
-        host_provider: host_provider.clone(),
+        host_provider,
         zenith,
         client: reqwest::Client::new(),
         sequencer_signer,
@@ -44,7 +43,7 @@ async fn main() -> eyre::Result<()> {
 
     let authenticator_jh = authenticator.spawn();
     let (submit_channel, submit_jh) = submit.spawn();
-    let build_jh = builder.spawn(submit_channel, host_provider.clone());
+    let build_jh = builder.spawn(submit_channel, ru_provider);
 
     let port = config.builder_port;
     let server = serve_builder_with_span(([0, 0, 0, 0], port), span);
--- a/bin/builder.rs
+++ b/bin/builder.rs
@@ -43,7 +43,7 @@ async fn main() -> eyre::Result<()> {
 
     let authenticator_jh = authenticator.spawn();
     let (submit_channel, submit_jh) = submit.spawn();
-    let build_jh = builder.spawn(submit_channel, ru_provider);
+    let build_jh = builder.spawn(submit_channel);
 
     let port = config.builder_port;
     let server = serve_builder_with_span(([0, 0, 0, 0], port), span);
--- a/src/config.rs
+++ b/src/config.rs
@@ -17,6 +17,7 @@ const HOST_RPC_URL: &str = "HOST_RPC_URL";
 const ROLLUP_RPC_URL: &str = "ROLLUP_RPC_URL";
 const TX_BROADCAST_URLS: &str = "TX_BROADCAST_URLS";
 const ZENITH_ADDRESS: &str = "ZENITH_ADDRESS";
+const BUILDER_HELPER_ADDRESS: &str = "BUILDER_HELPER_ADDRESS";
 const QUINCEY_URL: &str = "QUINCEY_URL";
 const BUILDER_PORT: &str = "BUILDER_PORT";
 const SEQUENCER_KEY: &str = "SEQUENCER_KEY"; // empty (to use Quincey) OR AWS key ID (to use AWS signer) OR raw private key (to use local signer)
@@ -33,7 +34,6 @@ const OAUTH_CLIENT_ID: &str = "OAUTH_CLIENT_ID";
 const OAUTH_CLIENT_SECRET: &str = "OAUTH_CLIENT_SECRET";
 const OAUTH_AUTHENTICATE_URL: &str = "OAUTH_AUTHENTICATE_URL";
 const OAUTH_TOKEN_URL: &str = "OAUTH_TOKEN_URL";
-const BUILDER_HELPER_ADDRESS: &str = "BUILDER_HELPER_ADDRESS";
 
 /// Configuration for a builder running a specific rollup on a specific host
 /// chain.
--- a/src/config.rs
+++ b/src/config.rs
@@ -33,6 +33,7 @@ const OAUTH_CLIENT_ID: &str = "OAUTH_CLIENT_ID";
 const OAUTH_CLIENT_SECRET: &str = "OAUTH_CLIENT_SECRET";
 const OAUTH_AUTHENTICATE_URL: &str = "OAUTH_AUTHENTICATE_URL";
 const OAUTH_TOKEN_URL: &str = "OAUTH_TOKEN_URL";
+const BUILDER_HELPER_ADDRESS: &str = "0x0000000000000000000000000000000000000000";
 
 /// Configuration for a builder running a specific rollup on a specific host
 /// chain.
@@ -50,6 +51,8 @@ pub struct BuilderConfig {
     pub tx_broadcast_urls: Vec<Cow<'static, str>>,
     /// address of the Zenith contract on Host.
     pub zenith_address: Address,
+    /// address of the Builder Helper contract on Host.
+    pub builder_helper_address: Address,
     /// URL for remote Quincey Sequencer server to sign blocks.
     /// Disregarded if a sequencer_signer is configured.
     pub quincey_url: Cow<'static, str>,
@@ -157,6 +160,7 @@ impl BuilderConfig {
                 .map(Into::into)
                 .collect(),
             zenith_address: load_address(ZENITH_ADDRESS)?,
+            builder_helper_address: load_address(BUILDER_HELPER_ADDRESS)?,
             quincey_url: load_url(QUINCEY_URL)?,
             builder_port: load_u16(BUILDER_PORT)?,
             sequencer_key: load_string_option(SEQUENCER_KEY),
--- a/src/config.rs
+++ b/src/config.rs
@@ -33,7 +33,7 @@ const OAUTH_CLIENT_ID: &str = "OAUTH_CLIENT_ID";
 const OAUTH_CLIENT_SECRET: &str = "OAUTH_CLIENT_SECRET";
 const OAUTH_AUTHENTICATE_URL: &str = "OAUTH_AUTHENTICATE_URL";
 const OAUTH_TOKEN_URL: &str = "OAUTH_TOKEN_URL";
-const BUILDER_HELPER_ADDRESS: &str = "0x0000000000000000000000000000000000000000";
+const BUILDER_HELPER_ADDRESS: &str = "BUILDER_HELPER_ADDRESS";
 
 /// Configuration for a builder running a specific rollup on a specific host
 /// chain.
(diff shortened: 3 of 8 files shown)


dylanlott:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/init4tech/builder/45?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#45** <a href="https://app.graphite.dev/github/pr/init4tech/builder/45?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/init4tech/builder/45?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#44** <a href="https://app.graphite.dev/github/pr/init4tech/builder/44?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#43** <a href="https://app.graphite.dev/github/pr/init4tech/builder/43?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/tasks/block.rs
```diff
@@ -155,12 +153,12 @@ impl BlockBuilder {
         }
     }
 
-    async fn _get_bundles(&mut self, in_progress: &mut InProgressBlock) {
+    async fn get_bundles(&mut self, in_progress: &mut InProgressBlock) {
         tracing::trace!("query bundles from cache");
         let bundles = self.bundle_poller.check_bundle_cache().await;
+        // OPTIMIZE: Sort bundles received from cache
```
Evalir, line 159: this is a good topic to discuss async/in other platform: what's gonna be the bundling strategy overall here? probably worth ticketing this.

src/tasks/block.rs
```diff
@@ -83,11 +83,9 @@ impl InProgressBlock {
 
         if let Ok(txs) = txs {
             self.unseal();
-            // extend the transactions with the decoded transactions.
-            // As this builder does not provide bundles landing "top of block", its fine to just extend.
```
Evalir, line 87: I would not remove these comments for now as we're still doing what the comment says, and serves as a reminder that this should probably change (else i'd also link the ticket about how to deal with bundles that i suggested below)

bin/builder.rs
```diff
@@ -38,7 +38,7 @@ async fn main() -> eyre::Result<()> {
 
     let authenticator_jh = authenticator.spawn();
     let (submit_channel, submit_jh) = submit.spawn();
-    let build_jh = builder.spawn(submit_channel);
+    let build_jh = builder.spawn(submit_channel, host_provider.clone());
```
anna-carroll, line 41: nit/question: is it necessary to clone it the second time?

src/config.rs
```diff
@@ -33,6 +33,7 @@ const OAUTH_CLIENT_ID: &str = "OAUTH_CLIENT_ID";
 const OAUTH_CLIENT_SECRET: &str = "OAUTH_CLIENT_SECRET";
 const OAUTH_AUTHENTICATE_URL: &str = "OAUTH_AUTHENTICATE_URL";
 const OAUTH_TOKEN_URL: &str = "OAUTH_TOKEN_URL";
+const BUILDER_HELPER_ADDRESS: &str = "BUILDER_HELPER_ADDRESS";
```
anna-carroll, line 36: super super nit: i like to keep everything in the same order. so i'd declare this constant right after ZENITH_ADDRESS

src/tasks/block.rs
```diff
@@ -155,14 +160,24 @@ impl BlockBuilder {
         }
     }
 
-    async fn _get_bundles(&mut self, in_progress: &mut InProgressBlock) {
+    /// Fetches bundles from the cache and ingests them into the in progress block
+    async fn get_bundles(&mut self, host_provider: &Provider, in_progress: &mut InProgressBlock) {
+        // Authenticate before fetching to ensure access to a valid token
+        if let Err(err) = self.bundle_poller.authenticator.authenticate().await {
```
anna-carroll, line 166: if this happens, do we want to continue building the block with regular transactions (as we're doing here) or do we want to fail more aggressively?

src/tasks/block.rs
```diff
@@ -205,9 +268,25 @@ impl BlockBuilder {
         self.secs_to_next_slot() + self.config.target_slot_time
     }
 
+    /// Parses bytes into a transaction envelope that is compatible with Flashbots-style bundles
+    fn parse_from_bundle(&self, tx: &Bytes) -> Result<(TxEnvelope, FixedBytes<32>), eyre::Error> {
```
anna-carroll, line 272: should this be a util / something that doesn't live in `block.rs`?

src/tasks/block.rs
```diff
@@ -172,6 +187,54 @@ impl BlockBuilder {
...
+    }
+
+    /// Simulates a transaction by calling it on the host provider at the current height with the current state.
+    async fn simulate_transaction(
+        &self,
+        host_provider: &Provider,
+        tx_env: TxEnvelope,
+    ) -> eyre::Result<()> {
+        let tx = TransactionRequest::from_transaction(tx_env);
+        host_provider.call(&tx).await?;
```
anna-carroll, line 234: huh. so does this throw an error if the `call` reverts?

anna-carroll, line 234: ❗ problem: isn't this a bundle of rollup transactions? this is simulating them against the host provider.

src/tasks/block.rs
```diff
@@ -172,6 +187,54 @@ impl BlockBuilder {
...
+        tracing::info!("simulating bundle");
+
+        let reverts = &bundle.bundle.reverting_tx_hashes;
+        tracing::debug!(reverts = ?reverts, "processing bundle with reverts");
+
+        for tx in &bundle.bundle.txs {
+            let (tx_env, hash) = self.parse_from_bundle(tx)?;
+
+            // Simulate and check for reversion allowance
+            match self.simulate_transaction(host_provider, tx_env).await {
```
anna-carroll, line 207: note: obviously, right now we're simulating all transactions against the current rollup state. we need to simulate a transaction, apply it, and simulate the next transaction against the chain state if the previous transaction is applied. e.g. generally more sophisticated simulation logic. this naive simulation logic may be "good enough" for testnet right now, but it doesn't give the assurances the Builder must actually give against reverting transactions

anna-carroll (changes requested):

src/tasks/block.rs
```diff
@@ -1,19 +1,23 @@
 use super::bundler::{Bundle, BundlePoller};
 use super::oauth::Authenticator;
 use super::tx_poller::TxPoller;
-use crate::config::{BuilderConfig, WalletlessProvider};
-use alloy::primitives::{keccak256, Bytes, B256};
-use alloy::providers::Provider;
+
```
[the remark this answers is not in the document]
dylanlott: No, but this did make me realize I had intended to organize the alloy import anyway, so I refactored that and cleaned it up 🧼

Evalir:
Will review first thing in the morning

src/tasks/block.rs
```diff
@@ -161,7 +159,11 @@ impl BlockBuilder {
...
 
     /// Fetches bundles from the cache and ingests them into the in progress block
-    async fn get_bundles(&mut self, host_provider: &Provider, in_progress: &mut InProgressBlock) {
+    async fn get_bundles(
+        &mut self,
+        ru_provider: &WalletlessProvider,
+        in_progress: &mut InProgressBlock,
+    ) {
         // Authenticate before fetching to ensure access to a valid token
         if let Err(err) = self.bundle_poller.authenticator.authenticate().await {
```
Evalir, line 168: the authenticator spawns a task which continuously refreshes the token. There is no need to programmatically do so.

src/tasks/block.rs
```diff
@@ -174,7 +176,7 @@ impl BlockBuilder {
         match bundles {
             Ok(bundles) => {
                 for bundle in bundles {
-                    let result = self.simulate_bundle(&bundle.bundle, host_provider).await;
+                    let result = self.simulate_bundle(&bundle.bundle, ru_provider).await;
```
Evalir, line 179: what are we simulating the bundle against? we don't have a simulation engine yet, nor we have a node we can use for this.

src/tasks/block.rs
```diff
@@ -187,13 +189,13 @@ impl BlockBuilder {
         self.bundle_poller.evict();
     }
 
-    /// Simulates a Flashbots-style ZenithEthBundle, simualating each transaction in it's bundle
+    /// Simulates a Flashbots-style `ZenithEthBundle`, simualating each transaction in its bundle
```
Evalir, line 192:
```suggestion
    /// Simulates a Flashbots-style `ZenithEthBundle`, simulating each transaction in its bundle
```

src/tasks/block.rs
```diff
@@ -224,14 +227,14 @@ impl BlockBuilder {
...
+    /// Simulates a rollup transaction by calling it on the ru provider at the current height with the current state.
     async fn simulate_transaction(
         &self,
-        host_provider: &Provider,
+        ru_provider: &WalletlessProvider,
         tx_env: TxEnvelope,
     ) -> eyre::Result<()> {
         let tx = TransactionRequest::from_transaction(tx_env);
-        host_provider.call(&tx).await?;
+        ru_provider.call(&tx).await?;
```
Evalir, line 237: so this is not entirely correct: bundles need to respect certain parameters they're sent with (which can be inspected by looking at the inner `EthSendBundle`) such as the block number, min/max timestamp, so on and so forth. some txs are also allowed to revert.

I'd say let's hold off from simulating and reconvene along with @prestwich on the sim engine to have an integration plan for this, as we cannot, at the moment, simulate bundles.

Evalir (changes requested):
I do not think we should be integrating a simulation flow right now, as we don't have all the necessary parts yet. If shelving the simulation engine means we're "blocked" on integrating simulation, we should discuss this on another avenue with @prestwich.

Evalir (changes requested):

src/tasks/block.rs
```diff
@@ -139,39 +141,104 @@ impl BlockBuilder {
...
+
+        trace!("query bundles from cache");
         let bundles = self.bundle_poller.check_bundle_cache().await;
+        // TODO: Sort bundles received from cache
         match bundles {
             Ok(bundles) => {
-                tracing::trace!("got bundles response");
                 for bundle in bundles {
-                    in_progress.ingest_bundle(bundle);
+                    let result = self.simulate_bundle(&bundle.bundle, ru_provider).await;
```
Evalir, line 179: alright, so as we talked yday I think let's just handwave simulation away from now. Once @prestwich has the sim engine ready, we can plug that in, whenever that happens. I wouldn't worry about this at all for now as it's introducing multiple caveats when simulating through `call`.

src/config.rs
```diff
@@ -17,6 +17,7 @@ const HOST_RPC_URL: &str = "HOST_RPC_URL";
 const ROLLUP_RPC_URL: &str = "ROLLUP_RPC_URL";
 const TX_BROADCAST_URLS: &str = "TX_BROADCAST_URLS";
 const ZENITH_ADDRESS: &str = "ZENITH_ADDRESS";
+const BUILDER_HELPER_ADDRESS: &str = "BUILDER_HELPER_ADDRESS";
```
Evalir, line 20: shall we loop in @rswanson for the new env var? or is this handled?

src/tasks/block.rs
```diff
@@ -139,39 +139,60 @@ impl BlockBuilder {
...
+                    if result.is_ok() {
+                        in_progress.ingest_bundle(bundle.clone());
+                    }
```
Evalir, line 174: we might want to log a simulation error if it's not OK (although this would ofc never be triggered rn)

Evalir (approved):
lgtm, looks great

src/config.rs
dylanlott: It's already handled and deployed 😃

anna-carroll:
> * Simulates those blocks, disallowing 4844 transactions and respecting the reverted transaction list

would update this PR comment @dylanlott

src/tasks/block.rs
```diff
@@ -139,39 +139,60 @@ impl BlockBuilder {
...
         self.bundle_poller.evict();
     }
 
+    /// Simulates a Zenith bundle against the rollup state
+    async fn simulate_bundle(
+        &mut self,
+        bundle: &ZenithEthBundle,
+        _ru_provider: &WalletlessProvider,
+    ) -> eyre::Result<()> {
+        debug!(hash = ?bundle.bundle.bundle_hash(), block_number = ?bundle.block_number(), "beginning bundle simulation");
```
anna-carroll, line 190: i might not log this right now as we're not actually doing it, but not a huge deal

src/tasks/block.rs
```diff
@@ -207,38 +228,108 @@ impl BlockBuilder {
...
                     // Build a block
                     let mut in_progress = InProgressBlock::default();
                     self.get_transactions(&mut in_progress).await;
-
-                    // TODO: Implement bundle ingestion #later
-                    // self.get_bundles(&mut in_progress).await;
+                    self.get_bundles(&ru_provider, &mut in_progress).await;
 
                     // Filter confirmed transactions from the block
                     self.filter_transactions(&mut in_progress).await;
```
anna-carroll, line 249: realizing right now that doing this line after ingesting bundles has the _potential_ to split up bundles. 

e.g. if someone submits txn A as an individual txn; shortly after, submits a bundle with txns A, B, C; the individual txn A gets picked up and included in a block; the bundle gets picked up in a subsequent block; this logic would pick up the bundle and filter out A while including B & C, thus splitting up the bundle.

in practice this is obviously an edge case. 

but keeping the code structure like this feels like a potential footgun 

 the `filter_transactions` function as it is currently implemented essentially intended to implement a simple simulation strategy for single transactions.

src/tasks/submit.rs
```diff
@@ -138,18 +145,19 @@ impl SubmitTask {
...
-        let header = Zenith::BlockHeader {
+        let header = zenith_types::BundleHelper::BlockHeader {
             hostBlockNumber: resp.req.host_block_number,
             rollupChainId: U256::from(self.config.ru_chain_id),
             gasLimit: resp.req.gas_limit,
             rewardAddress: resp.req.ru_reward_address,
             blockDataHash: in_progress.contents_hash(),
         };
 
+        let fills = vec![]; // NB: ignored until fills are implemented
```
anna-carroll, line 156: musing for later - when we implement fills, i suppose they should be part of the `InProgressBlock` struct

anna-carroll (commented):
the simulation for individual transactions could be at odds with un-simulated bundle logic...

src/tasks/block.rs
anna-carroll: also, if we were to just remove the `filter_transactions` call, then we'd perpetually include both individual txns and bundles that have already mined until they time out in the cache

Cargo.toml
```diff
@@ -50,4 +50,7 @@ tracing-subscriber = "0.3.18"
 async-trait = "0.1.80"
 oauth2 = "4.4.2"
 metrics = "0.24.1"
-metrics-exporter-prometheus = "0.16.0"
\ No newline at end of file
+metrics-exporter-prometheus = "0.16.0"
+
+[dev-dependencies]
+httpmock = "0.7.0"
```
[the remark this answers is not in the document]
dylanlott: Strange, I wrote a unit test with it somewhere, but it seems to have been clobbered out of git, I must investigate 🤔

src/tasks/block.rs
dylanlott: Right, the transaction reaping is necessary, but the above would invalidate the all-or-nothing aspect of the bundle, so then instead should the submit task should:

1. pull bundles & ingest them
2. pull transactions, ingest them
2.1. filtering txs against anything already in the bundles
3. Sign and submit the block like normal 

This would prevent bundle splitting and is still the correct pattern after bundle simulation is added.

src/tasks/block.rs
dylanlott: Can't use the `?` sugar here because it doesn't return a Result type, following the convention of the other functions.

anna-carroll (approved):

PR #48, opened 2025-01-15
feat: collect metrics on submitBlock transaction revert/success

anna-carroll:


src/tasks/receipts.rs
```diff
@@ -0,0 +1,74 @@
+use crate::config::Provider;
+use alloy::{primitives::TxHash, providers::Provider as _};
+use metrics::{counter, histogram};
+use std::time::Instant;
+use tokio::{sync::mpsc, task::JoinHandle};
+use tracing::{debug, error};
+
+/// Submits sidecars in ethereum txns to mainnet ethereum
```
dylanlott, line 8: I think this comment is a mistaken holdover from `SubmitTask` struct, should be updated to describe `ReceiptTask`.

src/tasks/receipts.rs
```diff
@@ -0,0 +1,74 @@
...
+#[derive(Debug, Clone)]
+pub struct ReceiptTask {
+    /// Ethereum Provider
+    pub host_provider: Provider,
+}
+
+impl ReceiptTask {
+    pub async fn log_tx(&self, pending_tx_hash: TxHash) {
```
dylanlott, line 16: nit: let's add a comment to this since it's a public function

src/tasks/mod.rs
```diff
@@ -1,5 +1,6 @@
 pub mod block;
 pub mod bundler;
 pub mod oauth;
+pub mod receipts;
```
dylanlott, line 4: super nit: this mod name is plural, but the rest are in the singular.

dylanlott (changes requested):
Just a few nits, but overall looks pretty good! 

We better get the sickest new grafana dashboard out of this 😤

src/tasks/receipts.rs
```diff
@@ -0,0 +1,74 @@
...
+        // start timer when tx hash is received
+        let start: Instant = Instant::now();
+
+        // wait for the tx to mine, get its receipt
+        let receipt = self.host_provider.clone().get_transaction_receipt(pending_tx_hash).await;
+
+        match receipt {
+            Ok(receipt) => {
```
Evalir, line 24: mostly a pedantic naming change, to make clear that this is a none type that can be none in the instance the tx simply isn't found.
```suggestion
            Ok(maybe_receipt) => {
```

src/tasks/receipts.rs
```diff
@@ -0,0 +1,74 @@
...
+                match receipt {
```
Evalir, line 25:
```suggestion
                match maybe_receipt {
```

src/tasks/receipts.rs
```diff
@@ -0,0 +1,74 @@
...
+        let (sender, mut inbound) = mpsc::unbounded_channel();
+        let handle = tokio::spawn(async move {
+            loop {
+                if let Some(pending_tx_hash) = inbound.recv().await {
+                    let this = self.clone();
+                    tokio::spawn(async move {
+                        let that = this.clone();
+                        that.log_tx(pending_tx_hash).await;
+                    });
+                } else {
```
Evalir, line 65: let's add some more trace/debug logging here so we can easily inspect when 1. the task is spawned 2. metric subtask are spawned (and "done" after the `.await` on line 63)

Evalir (changes requested):

src/tasks/receipts.rs
anna-carroll: https://github.com/init4tech/builder/pull/48/commits/7ebd8d219556a9fa66d2c3c697b445c254d6029c

src/tasks/receipts.rs
anna-carroll: https://github.com/init4tech/builder/pull/48/commits/7ebd8d219556a9fa66d2c3c697b445c254d6029c

src/tasks/mod.rs
anna-carroll: i ended up renaming the whole thing, i think makes better sense to call it `metrics`, and i kept it plural because `metric` sounds weird https://github.com/init4tech/builder/pull/48/commits/7ebd8d219556a9fa66d2c3c697b445c254d6029c

src/tasks/receipts.rs
anna-carroll: https://github.com/init4tech/builder/pull/48/commits/7ebd8d219556a9fa66d2c3c697b445c254d6029c

src/tasks/receipts.rs
anna-carroll: good catch https://github.com/init4tech/builder/pull/48/commits/7ebd8d219556a9fa66d2c3c697b445c254d6029c

Evalir (approved):
lgtm

PR #63, opened 2025-03-17
[wip] adds simulation factory to builder

dylanlott:
# [wip] Adds Simulator to Builder

This PR adds the SimulatorFactory to the Builder.

- adds initial implementation of simulator task to the builder
- adds trevm @ 0.20.x

# Related 
Closes ENG-790
Closes ENG-957

PR #69, opened 2025-04-09
breaking up simulator code

prestwich:
based on #63


src/tasks/sim/best.rs
```diff
@@ -0,0 +1,40 @@
+use alloy::primitives::U256;
+
+/// An evalutated EVM transaction with a particular score.
```
Evalir, line 3:
```suggestion
/// An evaluated EVM transaction with a particular score.
```

src/tasks/sim/factory.rs
```diff
@@ -0,0 +1,114 @@
...
+        Self { db, constants, execution_timeout, _pd: PhantomData }
+    }
+}
+
+impl<Db, Insp> DbConnect for SimFactory<Db, Insp>
+where
+    Db: DatabaseRef + Clone + Sync,
+    Insp: Sync,
+{
+    type Database = CacheOnWrite<Db>;
```
dylanlott, line 46: Why the change from Arc<ConcurrentState<Db>> to only CacheOnWrite<Db>?

src/tasks/sim/task.rs
```diff
@@ -0,0 +1,58 @@
...
+                handles.push(jh);
+            }
+
+            let mut best = None;
+            let best_score = U256::ZERO;
+
+            for handle in handles {
+                let result = handle.join().unwrap();
+                if let Some(candidate) = result {
+                    if *candidate.score() > best_score {
```
anna-carroll, line 49: note: if the scores for all transactions are zero then `best` will remain `None` 

(this might be what we want the behavior to be, or we might prefer that it returns one of the transactions?)

PR #71, opened 2025-04-15
[feat] adds simulation to block loop

dylanlott:
# [feat] adds simulation to block loop

This PR adds simulation to the block builder loop.

- Updates the builder to use a `SlotCalculator` for managing slot timings. 
- Removes the tx and bundle poller from the block builder and instead leans into the actor model and feeds them in via channels to the simulation process.
- This PR relies on the [signet-sdk Sim Crate PR](https://github.com/init4tech/signet-sdk/pull/33/files)

## Configuration

Deploying this PR will require additional environment variables to be set 

- `CONCURRENCY_LIMIT` to limit how many concurrent simulations the block builder runs. 
- `START_TIMESTAMP` to set the starting timestamp of the chain for anchoring slot calculation.

##  Testing

This code is still being tested before final merge.

The current test plan is to

- Build a Docker image of the builder with this branch
- Update environment variables to include the above values
- Deploy the test image to Pecorino
- Add a bundle to the cache and see if it is picked up
- See what, if any, errors spring up
- Land blocks and see that the block building rate is not decreased with new timing updates


dylanlott:
* **#71** <a href="https://app.graphite.dev/github/pr/init4tech/builder/71?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/init4tech/builder/71?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/tasks/tx_poller.rs
```diff
@@ -1,39 +1,94 @@
...
+
+    /// Returns a new [`TxPoller`] with the given config and cache polling interval in milliseconds.
+    pub fn new_with_poll_interval_ms(config: &BuilderConfig, poll_interval_ms: u64) -> Self {
+        Self { config: config.clone(), client: Client::new(), poll_interval_ms }
     }
 
-    /// polls the tx-pool for unique transactions and evicts expired transactions.
-    /// unique transactions that haven't been seen before are sent into the builder pipeline.
+    /// Polls the transaction cache for transactions.
     pub async fn check_tx_cache(&mut self) -> Result<Vec<TxEnvelope>, Error> {
```
anna-carroll, line 43: FYI, [this PR](https://github.com/init4tech/signet-sdk/pull/39) pulls this functionality into the SDK - but no need to block on it

src/tasks/bundler.rs
```diff
@@ -1,58 +1,62 @@
...
+        Self { config: config.clone(), authenticator, poll_interval_ms }
+    }
 
+    /// Fetches bundles from the transaction cache and returns them.
+    pub async fn check_bundle_cache(&mut self) -> eyre::Result<Vec<Bundle>> {
         let bundle_url: Url = Url::parse(&self.config.tx_pool_url)?.join("bundles")?;
         let token = self.authenticator.fetch_oauth_token().await?;
 
-        // Add the token to the request headers
         let result = reqwest::Client::new()
```
prestwich, line 60: rather than making a new client each time this is called, let's get https://github.com/init4tech/signet-sdk/pull/39 in and use the `TxCache` struct there

then the cache becomes a prop on `BundlePoller { cach: TxCache, .. }` and we avoid setting this up every time

prestwich (changes requested):
we are going places 😎

src/config.rs
```diff
@@ -116,6 +123,9 @@ pub enum ConfigError {
     /// Error connecting to the signer
     #[error("failed to connect to signer: {0}")]
     Signer(#[from] SignerError),
+    /// Error parsing the provided genesis file
+    #[error("failed")]
+    Genesis(String),
```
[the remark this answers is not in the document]
dylanlott: Good catch, this was dead code leftover from a refactor - removed. ❌

src/config.rs
```diff
@@ -93,6 +96,10 @@ pub struct BuilderConfig {
     pub oauth_token_url: String,
     /// The oauth token refresh interval in seconds.
     pub oauth_token_refresh_interval: u64,
+    /// The max number of simultaneous block simulations to run.
+    pub concurrency_limit: usize,
+    /// The anchor for slot time and number calculations before adjusting for chain offset.
+    pub start_timestamp: u64,
```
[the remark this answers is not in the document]
dylanlott: Yes, there is a ticket - ENG-1012 covers the config updates. 😃

src/tasks/tx_poller.rs
```diff
@@ -1,39 +1,94 @@
...
 
 /// Implements a poller for the block builder to pull transactions from the transaction pool.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct TxPoller {
-    /// config for the builder
+    /// Config values from the Builder.
     pub config: BuilderConfig,
-    /// Reqwest client for fetching transactions from the tx-pool
+    /// Reqwest Client for fetching transactions from the cache.
```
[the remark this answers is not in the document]
dylanlott: Will address in a follow up PR 👍

src/tasks/tx_poller.rs
dylanlott: Will address in a follow-up PR 👍

src/consts.rs
```diff
@@ -0,0 +1,4 @@
+//! Constants used in the builder.
```
Evalir, line 1: mega nit: let's just call this file `constants.rs` since that's the usual name elsewhere

src/config.rs
```diff
@@ -209,42 +211,71 @@ impl BuilderConfig {
...
+        );
+        let rollup = RollupConfig::new(
+            self.ru_chain_id,
+            address!("0x4E8cC181805aFC307C83298242271142b8e2f249"),
+            address!("0xd553C4CA4792Af71F4B61231409eaB321c1Dd2Ce"),
+            address!("0xe0eDA3701D44511ce419344A4CeD30B52c9Ba231"),
+            PredeployTokens::new(
+                address!("0x0B8BC5e60EE10957E0d1A0d95598fA63E65605e2"),
+                address!("0xF34326d3521F1b07d1aa63729cB14A372f8A737C"),
+                address!("0xE3d7066115f7d6b65F88Dff86288dB4756a7D733"),
```
Evalir, line 273: these should all be constants in `constants.rs`

src/config.rs
```diff
@@ -278,5 +309,18 @@ pub fn load_url(key: &str) -> Result<Cow<'static, str>, ConfigError> {
 /// Load an address from an environment variable.
 pub fn load_address(key: &str) -> Result<Address, ConfigError> {
     let address = load_string(key)?;
-    Address::from_str(&address).map_err(Into::into)
+    Address::from_str(&address)
+        .map_err(|_| ConfigError::Var(format!("Invalid address format for {}", key)))
```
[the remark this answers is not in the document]
prestwich: we're about to rip these out for `FromEnv` so i'd go ahead and skip hygiene here

src/tasks/block.rs
```diff
@@ -1,329 +1,352 @@
...
+    ///
+    /// - `tx_receiver`: A channel receiver for incoming transactions.
+    /// - `bundle_receiver`: A channel receiver for incoming bundles.
+    /// - `cache`: The simulation cache to store the received items.
+    ///
+    /// # Returns
+    ///
+    /// A `JoinHandle` for the basefee updater and a `JoinHandle` for the
+    /// cache handler.
+    pub fn spawn_cache_task(
```
prestwich, line 121: nit: this actually spawns 2 tasks

src/tasks/block.rs
```diff
@@ -1,329 +1,352 @@
...
+                return None;
             }
-            .in_current_span(),
-        )
+        };
+        let alloy_db: AlloyDB<Ethereum, RuProvider> =
+            AlloyDB::new(self.ru_provider.clone(), BlockId::from(latest));
+        let wrapped_db: WrapDatabaseAsync<AlloyDB<Ethereum, RuProvider>> =
+            WrapDatabaseAsync::new(alloy_db).unwrap_or_else(|| {
+                panic!("failed to acquire async alloy_db; check which runtime you're using")
```
prestwich, line 292: this errorr only occurs on current thread runtimes, so we can actually rule it out entirely by setting the runtime flavor in the main macro [link](https://docs.rs/tokio/latest/tokio/attr.main.html)

src/tasks/block.rs
```diff
@@ -1,329 +1,352 @@
...
+        finish_by: Instant,
+    ) -> Result<BuiltBlock, mpsc::error::SendError<BuiltBlock>> {
+        tracing::debug!(finish_by = ?finish_by, "starting block simulation");
+
+        let db = self.create_db().await.unwrap();
+        let block_build: BlockBuild<_, NoOpInspector> = BlockBuild::new(
+            db,
+            constants,
+            PecorinoCfg {},
+            NoopBlock,
```
prestwich, line 95: can't use a noop-block here. need to load a header and modify in a way similar to what happens here:

https://github.com/init4tech/signet-sdk/blob/main/crates/rpc/src/ctx.rs#L500-L523

prestwich (changes requested):

src/tasks/block.rs
```diff
@@ -1,329 +1,352 @@
...
+        cache: SimCache,
+    ) -> (JoinHandle<()>, JoinHandle<()>) {
+        tracing::debug!("starting up cache handler");
+
+        let basefee_price = Arc::new(AtomicU64::new(0_u64));
+        let basefee_reader = Arc::clone(&basefee_price);
+
+        // Update the basefee on a per-block cadence
+        let basefee_jh =
+            tokio::spawn(async move { self.basefee_updater(Arc::clone(&basefee_price)).await });
```
prestwich, line 134: this arc clone is unnecessary, as we already cloned above

src/tasks/block.rs
```diff
@@ -1,329 +1,352 @@
...
+            let sim_cache = cache.clone();
+            let finish_by = self.calculate_deadline();
+
+            match self.handle_build(constants, sim_cache, finish_by).await {
+                Ok(block) => {
+                    tracing::debug!(block = ?block, "built block");
+                    let _ = submit_sender.send(block);
+                }
+                Err(e) => {
+                    tracing::error!(err = %e, "failed to send block");
```
prestwich, line 256: failed to send, or failed to build?

src/tasks/block.rs
```diff
@@ -1,329 +1,352 @@
...
-        let current_slot_time = (curr_timestamp - self.config.chain_offset) % ETHEREUM_SLOT_TIME;
-        (ETHEREUM_SLOT_TIME - current_slot_time) % ETHEREUM_SLOT_TIME
-    }
+    /// Calculates the deadline for the current block simulation.
+    ///
+    /// # Returns
+    ///
+    /// An `Instant` representing the deadline.
+    pub fn calculate_deadline(&self) -> Instant {
+        let now = SystemTime::now();
```
prestwich, line 269: i don't like that we're constantly mixing timestamps and Duration/Instant. it makes it very hard to reason about the logic. cc @anna-carroll can we add more convenience functions to `SlotCalculator` to make these easier? what are the common actions we want to take using a calculator?

src/tasks/block.rs
```diff
@@ -1,329 +1,352 @@
...
+        let unix_seconds = now.duration_since(UNIX_EPOCH).expect("Time went backwards").as_secs();
 
-    // add a buffer to the beginning of the block slot.
-    fn secs_to_next_target(&self) -> u64 {
-        self.secs_to_next_slot() + self.config.target_slot_time
+        Instant::now().checked_add(Duration::from_secs(unix_seconds)).unwrap()
```
prestwich, line 272: `checked_add().unwrap()` is a little redundant

src/tasks/block.rs
```diff
@@ -1,329 +1,352 @@
...
-        let bundle = create_mock_bundle(&wallet).await;
-
-        // Save previous hash for comparison
-        let prev_hash = in_progress_block.contents_hash();
-
-        // Ingest the bundle
-        in_progress_block.ingest_bundle(bundle);
+/// Configuration struct for Pecorino network values.
+#[derive(Debug, Clone)]
+pub struct PecorinoCfg {}
```
prestwich, line 337: if no probs, use a simple structdef without the `{}`

```
pub struct PecorinoCfg;
```

prestwich: also should impl Copy

src/test_utils.rs
```diff
@@ -0,0 +1,72 @@
...
+};
+use eyre::Result;
+use std::str::FromStr;
+use tracing_subscriber::{EnvFilter, Layer, layer::SubscriberExt, util::SubscriberInitExt};
+
+/// Sets up a block builder with test values
+pub fn setup_test_config() -> Result<BuilderConfig> {
+    let config = BuilderConfig {
+        host_chain_id: 17000,
+        ru_chain_id: 17001,
```
prestwich, line 16: do these need to match the Pecorino chain id used below?

src/tasks/block.rs
dylanlott: Created [ENG-1019](https://linear.app/initiates/issue/ENG-1019/standardize-instant-and-duration-time-usage-in-slot-calculator) to capture this.

Cargo.toml
```diff
@@ -22,11 +22,24 @@ name = "transaction-submitter"
 path = "bin/submit_transaction.rs"
 
 [dependencies]
-init4-bin-base = "0.1.0"
+init4-bin-base = { git = "https://github.com/init4tech/bin-base.git" }
```
prestwich, line 25: can be changed back to "0.3"

src/tasks/block.rs
```diff
@@ -1,329 +1,515 @@
...
+
+        // Fetch the basefee from previous block to calculate gas for this block
+        let basefee = match self.get_basefee().await? {
+            Some(basefee) => basefee,
+            None => {
+                tracing::warn!("get basefee failed - RPC error likely occurred");
+                BASEFEE_DEFAULT
+            }
+        };
+        tracing::debug!(basefee = basefee, "setting basefee");
```
prestwich, line 358: nit don't need the `=` unless you're modifying something: 
`debug!(basefee, "setting basefee");`

src/tasks/block.rs
```diff
@@ -1,329 +1,515 @@
...
-        Self { transactions: Vec::new(), raw_encoding: OnceLock::new(), hash: OnceLock::new() }
-    }
 
-    /// Get the number of transactions in the block.
-    pub fn len(&self) -> usize {
-        self.transactions.len()
-    }
+/// Different error types that the Simulator handles
+#[derive(Debug, Error)]
+pub enum SimulatorError {
```
prestwich, line 47: type seems unnecessary for now?

prestwich (commented):
i think we should merge this today and then iterate

prestwich (approved):

PR #94, opened 2025-05-22
fix: submit channel refactors

dylanlott:
# fix: submit channel refactors

This PR includes several fixes and refactors to the channel submission logic. 

- Handle 403 errors with `Skip`s instead of erroring
- Adds `retry_count` increments to some code paths that weren't previously ticking the count
- Adds per-slot limiting to the block building loop so that the retry function doesn't simply get called again after 3 attempts
- Explicitly logs the `call` results of the blob transaction to aid with debugging
- Adds gas bumping logic to the submit task to facilitate transaction replacement

Additionally, there are some configuration changes that needed to be made.

- Additional broadcast hosts was just pointing back at the host RPC, which was adding noise to the logs and interfering with nonce debugging. This PR makes it optional and un-sets the duplicate host-rpc URL.
- Slot offset and start timestamp were incorrectly configured and have been updated to their correct values

Closes ENG-1079

See: [Builder Retry Logic](https://www.notion.so/init4tech/Builder-Retry-Logic-1f43e1cd45398092a1b3fb692dd37787?pvs=4)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,11 +27,11 @@ integration = []
 [dependencies]
 init4-bin-base = { version = "0.4.1", features = ["perms"] }
 
-signet-constants = { git = "https://github.com/init4tech/signet-sdk", rev = "bd183b627dcb0eb682da801093b13f1f8311446b" }
-signet-sim = { git = "https://github.com/init4tech/signet-sdk", rev = "bd183b627dcb0eb682da801093b13f1f8311446b" }
-signet-tx-cache = { git = "https://github.com/init4tech/signet-sdk", rev = "bd183b627dcb0eb682da801093b13f1f8311446b" }
-signet-types = { git = "https://github.com/init4tech/signet-sdk", rev = "bd183b627dcb0eb682da801093b13f1f8311446b" }
-signet-zenith = { git = "https://github.com/init4tech/signet-sdk", rev = "bd183b627dcb0eb682da801093b13f1f8311446b" }
+signet-constants = { git = "https://github.com/init4tech/signet-sdk", rev = "ba5894f6fac35299d495ae115cd465a1f0791637" }
+signet-sim = { git = "https://github.com/init4tech/signet-sdk", rev = "ba5894f6fac35299d495ae115cd465a1f0791637" }
+signet-tx-cache = { git = "https://github.com/init4tech/signet-sdk", rev = "ba5894f6fac35299d495ae115cd465a1f0791637" }
+signet-types = { git = "https://github.com/init4tech/signet-sdk", rev = "ba5894f6fac35299d495ae115cd465a1f0791637" }
+signet-zenith = { git = "https://github.com/init4tech/signet-sdk", rev = "ba5894f6fac35299d495ae115cd465a1f0791637" }
 
 trevm = { version = "0.23.4", features = ["concurrent-db", "test-utils"] }
 
--- a/src/tasks/submit.rs
+++ b/src/tasks/submit.rs
@@ -4,7 +4,7 @@ use crate::{
     utils::extract_signature_components,
 };
 use alloy::{
-    consensus::SimpleCoder,
+    consensus::{SimpleCoder, constants::GWEI_TO_WEI},
     eips::BlockNumberOrTag,
     network::{TransactionBuilder, TransactionBuilder4844},
     primitives::{FixedBytes, TxHash, U256},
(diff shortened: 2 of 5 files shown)


dylanlott:
* **#94** <a href="https://app.graphite.dev/github/pr/init4tech/builder/94?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/init4tech/builder/94?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Cargo.toml
```diff
@@ -27,11 +27,17 @@ integration = []
...
+# signet-sim = { path = "../signet-sdk/crates/sim" }
+# signet-tx-cache = { path = "../signet-sdk/crates/tx-cache" }
+# signet-types = { path = "../signet-sdk/crates/types" }
+# signet-zenith = { path = "../signet-sdk/crates/zenith" }
+
+signet-constants = { git = "https://github.com/init4tech/signet-sdk", branch = "dylan/block-number" }
+signet-sim = { git = "https://github.com/init4tech/signet-sdk", branch = "dylan/block-number" }
+signet-tx-cache = { git = "https://github.com/init4tech/signet-sdk", branch = "dylan/block-number" }
+signet-types = { git = "https://github.com/init4tech/signet-sdk", branch = "dylan/block-number" }
+signet-zenith = { git = "https://github.com/init4tech/signet-sdk", branch = "dylan/block-number" }
```
dylanlott, line 40: Will be set back to main after PR is merged.

src/tasks/submit.rs
```diff
@@ -53,17 +61,14 @@ pub enum ControlFlow {
 }
 
 /// Submits sidecars in ethereum txns to mainnet ethereum
-#[derive(Debug)]
```
prestwich, line 56: there should never be 2 of these, if we're cloning it something is wrong. clone impl ought to be removed

src/tasks/submit.rs
```diff
@@ -91,98 +96,162 @@ impl SubmitTask {
...
         {
-            error!(
-                code = e.code,
-                message = %e.message,
-                data = ?e.data,
-                "error in transaction submission"
-            );
-
+            // NB: These errors are all handled the same way but are logged for debugging purposes
             if e.as_revert_data()
```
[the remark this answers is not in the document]
prestwich: started messing around here
https://github.com/init4tech/builder/pull/98/files

Cargo.toml
```diff
@@ -25,17 +25,17 @@ path = "bin/submit_transaction.rs"
...
 [dependencies]
-init4-bin-base = { version = "0.3.4", features = ["perms"] }
+init4-bin-base = { version = "0.4.1", features = ["perms"] }
 
-signet-constants = { git = "https://github.com/init4tech/signet-sdk", rev = "b8251ff0fec7cb14ca87e6f95c14f56bc2593049" }
-signet-sim = { git = "https://github.com/init4tech/signet-sdk", rev = "b8251ff0fec7cb14ca87e6f95c14f56bc2593049" }
-signet-tx-cache = { git = "https://github.com/init4tech/signet-sdk", rev = "b8251ff0fec7cb14ca87e6f95c14f56bc2593049" }
-signet-types = { git = "https://github.com/init4tech/signet-sdk", rev = "b8251ff0fec7cb14ca87e6f95c14f56bc2593049" }
-signet-zenith = { git = "https://github.com/init4tech/signet-sdk", rev = "b8251ff0fec7cb14ca87e6f95c14f56bc2593049" }
+signet-constants = { git = "https://github.com/init4tech/signet-sdk", rev = "ba5894f6fac35299d495ae115cd465a1f0791637" }
```
Evalir, line 30: note that technically there's a version mismatch here: `bin-base` will pull `signet-constants` from commit `bd183b627dcb0eb682da801093b13f1f8311446b` while this dep will be exclusively using this commit. This shouldn't cause problems right now, but it is something to be aware of. We could bump signet-constants as a patch release to make sure it's at the commit? @prestwich

src/tasks/submit.rs
```diff
@@ -22,11 +22,19 @@ use signet_sim::BuiltBlock;
...
+use std::time::{Instant, UNIX_EPOCH};
 use tokio::{sync::mpsc, task::JoinHandle};
 
+// NB: Consider pulling the below gas parameters out into env vars
+/// Base maximum fee per gas to use as a starting point for retry bumps
+pub const BASE_FEE_PER_GAS: u128 = 10_000_000_000; // 10 Gwei
+/// Base max priority fee per gas to use as a starting point for retry bumps
+pub const BASE_MAX_PRIORITY_FEE_PER_GAS: u128 = 2_000_000_000; // 2 Gwei
+/// Base maximum fee per blob gas to use as a starting point for retry bumps
+pub const BASE_MAX_FEE_PER_BLOB_GAS: u128 = 1_000_000_000; // 1 Gwei
```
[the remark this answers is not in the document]
dylanlott: Addressed in #100 which passes a `BlockEnv` to the `SubmitTask` so that it has basefee info.

PR #95, opened 2025-05-23
chore(deps): pin signet-sdk to commit before larger update

Evalir:
For the current PRs to work without a huge rebase, we need to pin signet-sdk to the commit before what's in main, to use the old deps.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,12 +27,12 @@ integration = []
 [dependencies]
 init4-bin-base = "0.3"
 
-signet-bundle = { git = "https://github.com/init4tech/signet-sdk", branch = "main" }
-signet-constants = { git = "https://github.com/init4tech/signet-sdk", branch = "main" }
-signet-sim = { git = "https://github.com/init4tech/signet-sdk", branch = "main" }
-signet-tx-cache = { git = "https://github.com/init4tech/signet-sdk", branch = "main" }
-signet-types = { git = "https://github.com/init4tech/signet-sdk", branch = "main" }
-signet-zenith = { git = "https://github.com/init4tech/signet-sdk", branch = "main" }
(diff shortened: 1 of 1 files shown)


dylanlott:
My PR depends on a branch in the SDK [`dylan/block-number`](https://github.com/init4tech/signet-sdk/pull/59) though, so the pin will need to contain those changes.

dylanlott (changes requested):
🎀 Graphite made me write something 🦀

Evalir:
Yeah, looking at the open PRs, i think it's best to:
- merge this PR as is
- merge prestwich's PRs
- update to newer sdk/alloy/trevm/reth
- merge your PR

dylanlott:
My PR is stacked on @prestwich PRs already, so it would be better to stack the `builder` changes for `revm` / `alloy` on top of my submit refactors. 

So instead we should:
- Close this PR
- Merge the env and cache integration PRs
- Merge my submit refactors, keeping the `signet-sdk` dep pointed at `dylan/block-number`
- Update `alloy` / `revm` versions, and fix the bugs that incurs and then set the sdk branch back to `main`

This will allow `alloy` and `revm` update work to occur simultaneously with ongoing submit and simulation debugging.

prestwich:
`main` is currently broken. so fixing main is the first priority. I vote for merging this, then restacking the open PRs

prestwich:
to be clearer, this PR maintains the status quo, ie. It explicitly depends on a pre-change sdk commit. So it won't require any changes to open PRs

my preferred is

- merge this
- all open prs
- updates

Evalir:
ya my point with this pr is to restore status quo—we don't want to merge the ongoing work with red CI

prestwich (approved):

prestwich:
merging this to unblock the current PRs

PR #99, opened 2025-05-29
chore(deps): upgrade to latest alloy/sdk/trevm

Evalir:
Closes ENG-1109

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,17 +25,17 @@ path = "bin/submit_transaction.rs"
 integration = []
 
 [dependencies]
-init4-bin-base = { version = "0.3.4", features = ["perms"] }
+init4-bin-base = { version = "0.4.1", features = ["perms"] }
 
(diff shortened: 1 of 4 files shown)


Evalir:
wonder if we should explicitly switch to the `SimpleNonceManager` instead of the `CachedNonceManager` as we did on the node?

prestwich:
we honestly want custom nonce management behavior here. so should drop it out entirely? or write our own?

Evalir:
i think we probably want to write our own, but to keep this moving, i propose:
- switch to the previous nonce manager to keep the status quo
- merge
- write a custom nonce manager in another pr

prestwich:
> I rebased my dylan/submit-channel-refactors branch on top of this to test it, and it compiled and ran with a cargo update && cargo build with the confirmed new alloy and trevm versions, so I'm in favor of merging now and following up nonce management with the submit refactors.

this sounds real good to me

PR #100, opened 2025-06-04
feat: passes sim result to the submit tasks

dylanlott:
# feat: Passes SimResult to the submit task

This PR adds a `SimResult` type that wraps a `BuiltBlock` and a `BlockEnv` together and sends them both to the `SubmitTask` so that a block's environment is known at submission time for gas calculation purposes.

- Adds a `SimResult` type that binds a `BlockEnv` to a `BuiltBlock`
- Passes that `SimResult` to the `SubmitTask` for gas calculation purposes
- Adds some misc. logging and cleanup

## Bugfixes

- Takes cache reference and calculates simulation deadline _after_ the block environment has changed, since that ticks along the loop

Closes ENG-1046

### DIFF
--- a/bin/submit_transaction.rs
+++ b/bin/submit_transaction.rs
@@ -1,3 +1,5 @@
+//! A simple transaction submitter that sends a transaction to a recipient address
+//! on a regular interval for the purposes of roughly testing rollup mining.
 use alloy::{
     network::{EthereumWallet, TransactionBuilder},
     primitives::{Address, U256},
@@ -67,18 +69,29 @@ async fn main() {
     }
 }
 
+/// Sends a transaction to the specified recipient address
 async fn send_transaction(provider: &HostProvider, recipient_address: Address) {
     // construct simple transaction to send ETH to a recipient
+    let nonce = match provider.get_transaction_count(provider.default_signer_address()).await {
+        Ok(count) => count,
+        Err(e) => {
+            error!(error = ?e, "failed to get transaction count");
+            return;
+        }
+    };
+
     let tx = TransactionRequest::default()
         .with_from(provider.default_signer_address())
         .with_to(recipient_address)
         .with_value(U256::from(1))
+        .with_nonce(nonce)
         .with_gas_limit(30_000);
 
     // start timer to measure how long it takes to mine the transaction
     let dispatch_start_time: Instant = Instant::now();
 
     // dispatch the transaction
+    debug!(?tx.nonce, "sending transaction with nonce");
     let result = provider.send_transaction(tx).await.unwrap();
 
     // wait for the transaction to mine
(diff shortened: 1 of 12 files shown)


bin/submit_transaction.rs
```diff
@@ -95,10 +108,13 @@ async fn send_transaction(provider: &HostProvider, recipient_address: Address) {
...
 
-    let hash = receipt.transaction_hash.to_string();
+    record_metrics(dispatch_start_time, receipt);
+}
 
-    // record metrics for how long it took to mine the transaction
+/// Record metrics for how long it took to mine the transaction
+fn record_metrics(dispatch_start_time: Instant, receipt: alloy::rpc::types::TransactionReceipt) {
     let mine_time = dispatch_start_time.elapsed().as_secs();
+    let hash = receipt.transaction_hash.to_string();
```
prestwich, line 117: nit: better not to bind the var. can just inline it on the trace

```
debug!(
    success = receipt.status(), 
    hash = %receipt.transaction_hash,
    mine_time, hash, 
    "transaction mined"
);
```

src/tasks/block/sim.rs
```diff
@@ -79,22 +89,34 @@ impl Simulator {
...
             self.config.concurrency_limit,
             sim_items,
             self.config.rollup_block_gas_limit,
         );
 
         let built_block = block_build.build().await;
-        debug!(block_number = ?built_block.block_number(), "finished building block");
+        debug!(
+            tx_count = built_block.tx_count(),
+            block_number = ?built_block.block_number(),
```
prestwich, line 117: i know this is preserved from earlier, but i think the `?` is unnecessary here?

src/tasks/env.rs
```diff
@@ -96,10 +95,10 @@ impl EnvTask {
                 }
             };
             span.record("number", previous.number);
-            debug!("retrieved latest block");
 
             let env = self.construct_block_env(&previous);
-            debug!(?env, "constructed block env");
+            debug!(block_number = ?env.number, env.basefee, "constructed latest block env");
```
prestwich, line 100: thing the `?` is unnecessary as number is a number

src/tasks/submit.rs
```diff
@@ -500,66 +508,83 @@ impl SubmitTask {
...
     }
 }
 
-// Returns gas parameters based on retry counts.
-fn calculate_gas_limits(
+/// Calculates gas parameters based on the block environment and retry count.
+fn calculate_gas(retry_count: usize, block_env: &BlockEnv) -> (u128, u128, u128) {
+    let fallback_blob_basefee = 500;
+
+    match block_env.blob_excess_gas_and_price {
```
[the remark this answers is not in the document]
dylanlott: Updated to add `SimEnv` type and pass along the previous host header, and used that in the gas calculation funcs instead.

src/tasks/cache/task.rs
```diff
@@ -6,7 +6,8 @@ use tokio::{
     sync::{mpsc, watch},
     task::JoinHandle,
 };
-use trevm::revm::context::BlockEnv;
+
```
prestwich, line 9: nit: no newline

prestwich (changes requested):
NACK on passing the header through the stack, header should be retrieved at the place it's needed

prestwich (approved):
looks like we're now passing an RU header through instead of host, which is not strictly necessary but seems fine

PR #123, opened 2025-07-02
fix: properly calculate deadline, use `chrono` for calculating timestamps

Evalir:
We were calculating the deadline by subtracting from an `Instant::now()`, which we shouldn't do. We now simply subtract the buffer from the remaining time in the slot and add the result to the current instant instead.

This also switches to using `chrono` for anything regarding timestamps instead of using `std`, for parity with the slot calculator. Everything else uses `Instant` or `Duration` already.

Closes ENG-1167
Closes ENG-1019

### DIFF
--- a/src/tasks/submit/task.rs
+++ b/src/tasks/submit/task.rs
@@ -171,7 +171,7 @@ impl SubmitTask {
                 .send_transaction(req)
                 .instrument(span.clone())
                 .await
-                .inspect_err(|e| error!(error = %e, "sending transaction"))
+                .inspect_err(|e| error!(error = %e, "error sending transaction"))
                 .unwrap_or(ControlFlow::Retry);
 
             let guard = span.entered();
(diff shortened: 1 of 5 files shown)


src/tasks/submit/task.rs
```diff
@@ -151,27 +157,26 @@ impl SubmitTask {
...
+            // Check at the top of the loop if the slot is still valid. This
+            // will prevent unnecessary retries if the slot has changed.
+            check_slot_still_valid!(self, expected_slot);
 
-            let guard = span.entered();
+            let inbound_result = self
+                .send_transaction(req)
+                .instrument(span.clone())
+                .await
+                .inspect_err(|e| error!(error = %e, "sending transaction"))
```
Evalir, line 174: this was probably written this way because it's implied with the error i assume, but i still think it can be confusing at first glance
```suggestion
                .inspect_err(|e| error!(error = %e, "error sending transaction"))
```

Evalir (commented):
lgtm, just one nit

PR #132, opened 2025-07-24
chore: misc improvements

Evalir:


src/tasks/cache/tx.rs
```diff
@@ -8,6 +8,9 @@ use serde::{Deserialize, Serialize};
 use std::time::Duration;
 use tokio::{sync::mpsc, task::JoinHandle, time};
 
+/// Poll interval for the transaction poller in milliseconds.
+const POLL_INTERVAL_MS: u64 = 1000;
```
dylanlott, line 12: Question: Why not an env var? If we changed it once here, it seems likely we'll want/need to change it again in the future, so it would likely be useful to have this as an optional env var instead of a hardcoded value.

Evalir: I didn't change this, I just made it a constant. I don't think it's also necessarily great to have this as a config value as it can be used to abuse polling the cache by reducing the poll time.

PR #137, opened 2025-08-06
refactor: use the provide config structs

prestwich:
Refactors builder config to use the new provider config structs

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -12,7 +12,6 @@ use alloy::{
             SimpleNonceManager, WalletFiller,
         },
     },
-    rpc::client::BuiltInConnectionString,
 };
 use eyre::Result;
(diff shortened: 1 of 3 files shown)


src/config.rs
```diff
@@ -68,14 +68,14 @@ pub struct BuilderConfig {
         var = "HOST_RPC_URL",
         desc = "URL for Host RPC node. This MUST be a valid HTTP or WS URL, starting with http://, https://, ws:// or wss://"
     )]
-    pub host_rpc_url: url::Url,
+    pub host_rpc: init4_bin_base::utils::provider::ProviderConfig,
```
Evalir, line 71: is there any reason we're using the entire path instead of just importing the used structs?

PR #148, opened 2025-09-10
feat(config): Set `SlotCalculator` config with `CHAIN_NAME` instead of verbose values

Evalir:
This makes it easier to configure the builder slot calculator through setting the chain name, instead of setting each slot calculator configuration variable.

Closes ENG-1372

@rswanson this will lets us load the slot calculator from the `CHAIN_NAME` env var.

This also updates signet dependencies because it's needed to use the latest `bin-base`. No behavior change is expected.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3319,9 +3319,9 @@ dependencies = [
 
 [[package]]
 name = "init4-bin-base"
-version = "0.12.2"
+version = "0.12.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9e3002542e3353a2cb915ce582df044c67936f2a6d8b8e3f1be490eb97d47d13"
+checksum = "9f69beaf3dde29492ac14a781de91655f745468703dde7bd5613222f11331f68"
 dependencies = [
  "alloy",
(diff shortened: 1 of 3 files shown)


Evalir:
* **#148** <a href="https://app.graphite.dev/github/pr/init4tech/builder/148?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/init4tech/builder/148?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

prestwich (changes requested):
This makes the builder unrunnable for any other chain. is that intentional?

Evalir:
It's runnable for any chain that the slot calculator supports through `KnownChains`? Which I think we're ok with?

prestwich:
this is a significant change to builder functionality that makes it impossible to run on custom nets. is that intentional?

Evalir:
Yes this is intentional. We want to make the builder easier to configure. Restricting the builders to chains we have the values in constants for is one way to do so.

prestwich (approved):

PR #151, opened 2025-09-17
chore: pass block-associated span with simenv and simresult

prestwich:


Evalir (approved):
very cool. really wanna see how this looks in prod

prestwich:
cc @dylanlott I think this causes minimal rebase issues with your PR?

PR #158, opened 2025-10-02
wip: adds a flashbots provider

dylanlott:


src/tasks/submit/flashbots.rs
```diff
@@ -108,37 +117,47 @@ impl FlashbotsTask {
...
+                .with_auth(self.signer.clone())
+                .await;
+
+            match response {
+                Ok(Some(hash)) => {
+                    tracing::info!(host_block_number = sim_result.host_block_number(), bundle_hash = ?hash, "bundle submitted to flashbots");
+                    span_debug!(span, "bundle submitted to flashbots");
+                }
+                Ok(None) => {
+                    tracing::info!(
```
prestwich: duplicate event

src/config.rs
```diff
@@ -173,14 +186,15 @@ pub struct BuilderConfig {
 
 impl BuilderConfig {
     /// Connect to the Builder signer.
-    pub async fn connect_builder_signer(&self) -> Result<LocalOrAws, SignerError> {
+    pub async fn connect_builder_signer(&self) -> eyre::Result<LocalOrAws> {
```
prestwich: why is this type changing?

dylanlott: To make it work with the error type `try_join` expects. The other `connect_*` functions return `eyre::Result` types, so it must match the others' return signatures.

prestwich (approved):

PR #159, opened 2025-10-07
fix glibc errors

dylanlott:


rswanson (commented):
What changed that impacted glibc btw, just curious

dylanlott:
Root cause was that 5 days ago, a new release of `cargo-chef`'s changed the version of `glibc` in the `lukemathwalker/cargo-chef` builds to a newer version of `glibc` than was compatible with the `builder` runtime stage 3, so builds on `main` branch were building but erroring at runtime. Changing to the `rust:bookworm` image aligns the `glibc` version with what's available in stage 3's `bookworm-slim` image.

Re: Image version - Version specification can loosen to `rust:1-bookworm` or even `rust:bookworm`. Those are much less strict and still solve the `glibc` issue and prevent it popping up in the future. I've updated it to `rust:bookworm` for now.

Evalir (approved):
perfect

PR #167, opened 2025-11-05
feat: adds host simulation to block build

dylanlott:
# feat: adds host simulation to block build

This PR adds host simulation to the block build loop.

> Note: Requires the signet-sdk release of `0.13.0`​ in order to pass CI.

- bumps bin-base to `0.15.1`​
- deprecates the `BuilderHelper`​ wrapper contract submission in favor of the raw `Zenith`​ contract
- wires up the `HostProvider`​ into the Simulator for host transaction simulation

### DIFF
--- a/src/tasks/env.rs
+++ b/src/tasks/env.rs
@@ -206,14 +206,35 @@ impl EnvTask {
         Self { config, host_provider, ru_provider }
     }
 
-    /// Construct a [`BlockEnv`] for the _next_ block from the previous block's [`Header`].
-    fn construct_env(&self, previous: Header) -> Environment {
+    /// Construct a [`BlockEnv`] for the next host block from the previous host header.
+    fn construct_host_env(&self, previous: Header) -> Environment {
         let env = BlockEnv {
             number: U256::from(previous.number + 1),
             beneficiary: self.config.builder_rewards_address,
-            // NB: EXACTLY the same as the previous block
+            // NB: EXACTLY the same as the previous block timestamp + slot duration
             timestamp: U256::from(previous.timestamp + self.config.slot_calculator.slot_duration()),
             gas_limit: self.config.max_host_gas(previous.gas_limit),
+            basefee: previous
+                .next_block_base_fee(BaseFeeParams::ethereum())
+                .expect("signet has no non-1559 headers"),
+            difficulty: U256::ZERO,
+            prevrandao: Some(B256::random()),
+            blob_excess_gas_and_price: Some(BlobExcessGasAndPrice {
+                excess_blob_gas: 0,
+                blob_gasprice: 0,
+            }),
+        };
+        Environment::new(env, previous)
+    }
+
+    /// Construct a [`BlockEnv`] for the next rollup block from the previous block header.
+    fn construct_rollup_env(&self, previous: Header) -> Environment {
+        let env = BlockEnv {
+            number: U256::from(previous.number + 1),
+            beneficiary: self.config.builder_rewards_address,
+            // NB: EXACTLY the same as the previous block timestamp + slot duration
+            timestamp: U256::from(previous.timestamp + self.config.slot_calculator.slot_duration()),
+            gas_limit: self.config.rollup_block_gas_limit,
             basefee: previous
                 .next_block_base_fee(BaseFeeParams::ethereum())
                 .expect("signet has no non-1559 headers"),
@@ -263,9 +284,9 @@ impl EnvTask {
             .header
             .inner;
 
-            // Construct the block envs using the previous block headers for each.
-            let rollup_env = self.construct_env(rollup_header.into());
-            let host_env = self.construct_env(host_header);
+            // Construct the block env using the previous block header
+            let rollup_env = self.construct_rollup_env(rollup_header.into());
+            let host_env = self.construct_host_env(host_header);
 
             span_debug!(
                 span,


src/tasks/submit/prep.rs
```diff
@@ -92,26 +92,32 @@ impl<'a> SubmitPrep<'a> {
...
-
-        let data = BundleHelper::submitCall { fills: self.fills(), header, v, r, s }.abi_encode();
+        debug!(?header.hostBlockNumber, "built zenith block header");
+
+        let data = Zenith::submitBlockCall {
+            header,
+            v,
+            r,
+            s,
+            _4: self
```
prestwich: this adds a bunch of calldata that should not be there

src/tasks/block/sim.rs
```diff
@@ -251,12 +269,19 @@ impl Simulator {
         deadline.max(Instant::now())
     }
 
+    /// Creates an `AlloyDB` instnace from the host provider.
```
anna-carroll: nit: typo: `instance`

src/tasks/block/sim.rs
```diff
@@ -127,19 +131,33 @@ impl Simulator {
         block_env: BlockEnv,
     ) -> eyre::Result<BuiltBlock> {
         let concurrency_limit = self.config.concurrency_limit();
+        let latest_block_number = BlockNumber::from(block_env.number.to::<u64>() - 1);
```
anna-carroll: nit: could call this `latest_ru_block_number` for explicit avoidance of doubt?

anna-carroll (commented):
nits

anna-carroll (commented):
need to leave keyboard for a sec but need to keep looking deeper at this code to determine what is supposed to be correct here. leaving these comments for now

(pretty much mirroring @Evalir 's callout above)

src/tasks/env.rs
```diff
@@ -70,9 +206,30 @@ impl EnvTask {
         Self { config, host_provider, ru_provider }
     }
 
-    /// Construct a [`BlockEnv`] by from the previous block header.
-    fn construct_block_env(&self, previous: &Header) -> BlockEnv {
-        BlockEnv {
+    /// Construct a [`BlockEnv`] by the previous host header.
+    fn construct_host_env(&self, previous: Header) -> Environment {
+        let env = BlockEnv {
+            number: U256::from(previous.number),
```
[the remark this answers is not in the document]
dylanlott: > rollup env is adding 1 to previous.number. host is just using previous.number. is this intentional?

It was previously done for testing, but the refactor made this incorrect as it handled the issue elsewhere.

> can we DRY here?

yes we can DRY it up now since the refactors addressed the DB target block issue. 👍

src/tasks/env.rs
```diff
@@ -70,9 +121,30 @@ impl EnvTask {
...
+            beneficiary: self.config.builder_rewards_address,
+            // NB: EXACTLY the same as the previous block
+            timestamp: U256::from(previous.timestamp + self.config.slot_calculator.slot_duration()),
```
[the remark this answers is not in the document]
dylanlott: Updated the comment to be more specific, but this makes an environment for the _next_ block.

src/tasks/env.rs
```diff
@@ -70,14 +206,14 @@ impl EnvTask {
...
 
-    /// Construct a [`BlockEnv`] by from the previous block header.
-    fn construct_block_env(&self, previous: &Header) -> BlockEnv {
-        BlockEnv {
+    /// Construct a [`BlockEnv`] for the _next_ block from the previous block's [`Header`].
+    fn construct_env(&self, previous: Header) -> Environment {
+        let env = BlockEnv {
             number: U256::from(previous.number + 1),
             beneficiary: self.config.builder_rewards_address,
             // NB: EXACTLY the same as the previous block
```
prestwich: this NB is incorrect?

src/tasks/env.rs
```diff
@@ -70,14 +206,14 @@ impl EnvTask {
...
             timestamp: U256::from(previous.timestamp + self.config.slot_calculator.slot_duration()),
-            gas_limit: self.config.rollup_block_gas_limit,
+            gas_limit: self.config.max_host_gas(previous.gas_limit),
```
prestwich: this is now incorrect for the rollup. they need different gas limits

prestwich (changes requested):

prestwich:
yeah its a bit unpleasant. could dry parts of it i think but other parts need to be separate

src/config.rs
```diff
@@ -292,4 +276,11 @@ impl BuilderConfig {
                 .unwrap_or(DEFAULT_CONCURRENCY_LIMIT)
         })
     }
+
+    /// Returns the maximum host gas to use for block building based on the configured max host gas coefficient.
+    pub fn max_host_gas(&self, gas_limit: u64) -> u64 {
+        // Set max host gas to a percentage of the host block gas limit
+        ((gas_limit as u128 * (self.max_host_gas_coefficient.unwrap_or(80) as u128)) / 100u128)
```
Evalir: ditto, this should be a constant so there's only 1 place that needs to change if we decide to change the coefficient

Evalir: yeah i'm happy with that, I think it's safer that way and makes other builders understand that changing this means you're running modified software

src/tasks/env.rs
```diff
@@ -70,12 +206,33 @@ impl EnvTask {
...
+    fn construct_host_env(&self, previous: Header) -> Environment {
+        let env = BlockEnv {
+            number: U256::from(previous.number + 1),
+            beneficiary: self.config.builder_rewards_address,
+            // NB: EXACTLY the same as the previous block timestamp + slot duration
+            timestamp: U256::from(previous.timestamp + self.config.slot_calculator.slot_duration()),
+            gas_limit: self.config.max_host_gas(previous.gas_limit),
+            basefee: previous
+                .next_block_base_fee(BaseFeeParams::ethereum())
+                .expect("signet has no non-1559 headers"),
```
anna-carroll: is this a bug for the Host chain?

prestwich (approved):

PR #169, opened 2025-11-07
fix: move block env creation for host to env task

prestwich:


src/tasks/block/sim.rs
```diff
@@ -58,6 +54,16 @@ pub struct SimResult {
 }
 
 impl SimResult {
+    /// Get a reference to the previous host header.
+    pub const fn prev_host(&self) -> &Header {
+        self.sim_env.prev_host()
+    }
+
+    /// Get a reference to the previous signet header.
```
anna-carroll: nit: `previous rollup header`

anna-carroll: thought: should we rename `prev_header` to `prev_rollup` to match `prev_host`?

anna-carroll: same question `block_number` to `rollup_block_number`?

anna-carroll (approved):
only naming and comments thoughts

Evalir (approved):

src/tasks/env.rs
```diff
@@ -71,8 +122,8 @@ impl EnvTask {
     }
 
     /// Construct a [`BlockEnv`] by from the previous block header.
-    fn construct_block_env(&self, previous: &Header) -> BlockEnv {
-        BlockEnv {
+    fn construct_block_env(&self, previous: Header) -> Environment {
+        let env = BlockEnv {
             number: U256::from(previous.number + 1),
```
anna-carroll: you can see here, the rollup block env was being created by adding 1 to the previous header's number. by refactoring to move the host block env creation to use the same function, we fix the bug

dylanlott (approved):

dylanlott:
Approved, we can merge this stack down into main and test for final release.

PR #171, opened 2025-11-10
fix: instantiate DB on previous block state

anna-carroll:


tests/block_builder_test.rs
```diff
@@ -45,13 +45,8 @@ async fn test_handle_build() {
...
 
-    let block_env = EnvTask::new(
-        config.clone(),
-        config.connect_host_provider().await.unwrap(),
-        ru_provider.clone(),
-    )
-    .spawn()
-    .0;
+    let block_env =
+        EnvTask::new(config.clone(), host_provider.clone(), ru_provider.clone()).spawn().0;
```
anna-carroll: this is a drive-by test fix to avoid connecting provider twice which wasn't necessary

src/tasks/block/sim.rs
```diff
@@ -304,11 +301,10 @@ impl Simulator {
     }
 
     /// Creates an `AlloyDB` instance from the host provider.
-    async fn create_host_db(&self) -> HostAlloyDatabaseProvider {
-        let block_height = self.host_provider.get_block_number().await.unwrap();
-        debug!(%block_height, "creating host alloyDB at block height");
+    async fn create_host_db(&self, latest_block_number: u64) -> HostAlloyDatabaseProvider {
```
prestwich: this fn does not need to be async anymore :)

PR #179, opened 2025-11-20
testing: adds a full test harness

dylanlott:
# testing: adds a full test harness

This PR adds a test harness to the builder. This test harness will facilitate deeper block lifecycle testing and assertions around smart contract calls and simulation between host and rollup chains. 

- adds a test harness with host and rollup Anvil chains 
- sets up initial lifecycle methods

### DIFF
--- a/tests/harness.rs
+++ b/tests/harness.rs
@@ -15,31 +15,37 @@ use std::{
 
 use alloy::{
     consensus::Header,
-    eips::BlockId,
-    network::Ethereum,
+    eips::{BlockId, eip1559::BaseFeeParams},
+    network::{Ethereum, EthereumWallet},
     node_bindings::Anvil,
-    primitives::U256,
-    providers::{Provider, RootProvider, ext::AnvilApi, layers::AnvilProvider},
+    primitives::{B256, U256},
+    providers::{
+        Provider, ProviderBuilder, RootProvider,
+        ext::AnvilApi,
+        fillers::{BlobGasFiller, SimpleNonceManager},
+        layers::AnvilProvider,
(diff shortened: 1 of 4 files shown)


tests/harness.rs
```diff
@@ -0,0 +1,307 @@
...
+            rollup_header_number = ru_header.number,
+            rollup_header_gas_limit = ru_header.gas_limit
+        );
+
+        let headers = (ru_header, host_header);
+        Ok(headers)
+    }
+
+    /// Tick a new `SimEnv` computed from the current latest rollup and host headers.
+    pub async fn tick_from_headers(&self, prev_ru_header: Header, prev_host_header: Header) {
```
prestwich: this is now duplicating env construction code from the env task. is there a reason we're not using an EnvTask here? it seems like we could easily have one running but control anvil to only produce WS notifications when a `tick` function is called

tests/harness.rs
```diff
@@ -0,0 +1,307 @@
...
+type RollupHeader = Header;
+type HostHeader = Header;
+type Headers = (RollupHeader, HostHeader);
+
+impl TestHarness {
+    /// Create a new harness with a fresh Anvil rollup chain and default test config.
+    pub async fn new() -> eyre::Result<Self> {
+        setup_logging();
+        let mut config = setup_test_config()?;
+        configure_slot_timing(&mut config)?;
```
prestwich: this is an anti-pattern. if we want the config to have a specific value, we should set it up that way. i;m not sure why we want to change this at all. we should be using either the test value in constants, or the pecorino value in constants

PR #180, opened 2025-11-21
chore(config): make flashbots URL not optional

Evalir:
No reason to have this optional, probably vestigial from when we had multiple submit tasks.


dylanlott:
Definitely vestigial, good to remove.

dylanlott (approved):

PR #185, opened 2025-11-25
chore(metrics): Refactor current metrics

Evalir:
We're not following our set pattern for metrics (see node-components) and it makes it hard to know what we're really tracking and where.

### DIFF
--- a/src/tasks/submit/flashbots.rs
+++ b/src/tasks/submit/flashbots.rs
@@ -14,6 +14,7 @@ use alloy::{
 };
 use eyre::OptionExt;
 use init4_bin_base::utils::signer::LocalOrAws;
+use std::time::Instant;
 use tokio::{sync::mpsc, task::JoinHandle};
 use tracing::{Instrument, debug, debug_span};
 
@@ -196,12 +197,16 @@ impl FlashbotsTask {
             let signer = self.signer.clone();
 
             tokio::spawn(async move {
+                let start = Instant::now();
                 let response = flashbots
                     .send_mev_bundle(bundle.clone())
                     .with_auth(signer.clone())
                     .into_future()
                     .instrument(submit_span.clone())
                     .await;
+
+                metrics::flashbots_submission_duration_ms()
+                    .record(start.elapsed().as_millis() as f64);
 
                 match response {
                     Ok(resp) => {
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -51,6 +51,10 @@ const FLASHBOTS_EMPTY_BLOCKS_HELP: &str = "Number of empty blocks skipped";
 const FLASHBOTS_SUBMISSIONS: &str = "signet.builder.flashbots.submissions";
 const FLASHBOTS_SUBMISSIONS_HELP: &str = "Number of submission attempts to Flashbots";
 
+const FLASHBOTS_SUBMISSION_DURATION_MS: &str = "signet.builder.flashbots.submission_duration_ms";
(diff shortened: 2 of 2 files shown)


src/tasks/block/sim.rs
```diff
@@ -158,8 +156,8 @@ impl Simulator {
             block_number = built_block.block_number(),
             "block simulation completed",
         );
-        counter!("signet.builder.built_blocks").increment(1);
-        histogram!("signet.builder.built_blocks.tx_count").record(built_block.tx_count() as u32);
+        metrics::built_blocks().increment(1);
+        metrics::built_blocks_tx_count().record(built_block.tx_count() as u32);
```
prestwich: these lines should both be collapsed into `metrcis::record_bulit_block(&block)` and the logic to increment, etc should be in the `metrics` file

src/tasks/submit/flashbots.rs
```diff
@@ -195,24 +197,28 @@ impl FlashbotsTask {
...
             tokio::spawn(async move {
+                let start = Instant::now();
                 let response = flashbots
                     .send_mev_bundle(bundle.clone())
                     .with_auth(signer.clone())
                     .into_future()
                     .instrument(submit_span.clone())
                     .await;
 
+                metrics::flashbots_submission_duration_ms()
```
prestwich: ditto

src/tasks/submit/flashbots.rs
```diff
@@ -195,24 +197,28 @@ impl FlashbotsTask {
...
+                        metrics::flashbots_bundles_submitted().increment(1);
                         span_debug!(
                             submit_span,
                             hash = resp.map(|r| r.bundle_hash.to_string()),
                             "received bundle hash after submitted to flashbots"
                         );
                     }
                     Err(err) => {
-                        counter!("signet.builder.flashbots.submission_failures").increment(1);
+                        metrics::flashbots_submission_failures().increment(1);
```
prestwich: ditto

src/tasks/metrics.rs
```diff
@@ -47,24 +46,24 @@ impl MetricsTask {
                 Ok(receipt) => {
                     // record how long it took to mine the transaction
                     // potential improvement: use the block timestamp to calculate the time elapsed
-                    histogram!("metrics.tx_mine_time").record(start.elapsed().as_millis() as f64);
+                    metrics::tx_mine_time_ms().record(start.elapsed().as_millis() as f64);
 
                     // log whether the transaction reverted
                     if receipt.status() {
-                        counter!("metrics.tx_succeeded").increment(1);
+                        metrics::tx_succeeded().increment(1);
```
prestwich: ditto

src/tasks/metrics.rs
```diff
@@ -47,24 +46,24 @@ impl MetricsTask {
...
                         debug!("tx succeeded");
                     } else {
-                        counter!("metrics.tx_reverted").increment(1);
+                        metrics::tx_reverted().increment(1);
```
prestwich: ditto

src/tasks/metrics.rs
```diff
@@ -47,24 +46,24 @@ impl MetricsTask {
...
                         debug!("tx reverted");
                     }
                 }
                 Err(PendingTransactionError::TxWatcher(WatchTxError::Timeout)) => {
                     // log that the transaction timed out
-                    counter!("metrics.tx_not_mined").increment(1);
+                    metrics::tx_not_mined().increment(1);
```
prestwich: ditto

PR #190, opened 2025-12-03
tests: adds a test for host and rollup transaction simulation

dylanlott:
# tests: adds a test for host and rollup transaction simulation

This PR adds a test that makes an anvil instance for the host and rollup provider and then uses that to simulate the block building loop. 

- Refactors the SimulateTask and EnvTask constructors to take Providers instead of privately setting them up for easier swapping at test time.

### DIFF
--- a/bin/builder.rs
+++ b/bin/builder.rs
@@ -16,17 +16,15 @@ async fn main() -> eyre::Result<()> {
     let init_span_guard = info_span!("builder initialization");
     let config = builder::config_from_env();
 
-    let (env_host_provider, sim_host_provider, env_ru_provider, sim_ru_provider, quincey) = tokio::try_join!(
+    let (host_provider, ru_provider, quincey) = tokio::try_join!(
         config.connect_host_provider(),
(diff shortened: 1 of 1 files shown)


tests/block_builder_test.rs
```diff
@@ -1,72 +1,195 @@
...
+        .keys()
+        .iter()
+        .take(2)
+        .map(|key| PrivateKeySigner::from_bytes(&B256::from_slice(&key.to_bytes())).unwrap())
+        .collect();
+    assert!(signers.len() >= 2, "rollup anvil must provide at least two accounts");
+    let provider = RootProvider::new_http(anvil.endpoint_url());
+
+    (anvil, provider, signers)
+}
```
dylanlott: NB: These are intentionally not forked off of RPC so that they don't rely on Pecorino state at all yet. These are simply simulating "dumb" Signet bundles. The next step to shore these up is going to be forking them off of the actual host and rollup RPCs (once Pecorino is running again) and then making these "smart" bundles that actually call Signet contracts.

PR #215, opened 2026-02-05
feat(submit): pylon blob preloading integration

Evalir:
This PR integrates Pylon blob sidecar preloading into the Flashbots submission flow:

  - Adds a PylonClient — new config field (PYLON_URL) and client for the Pylon blob server API. (cc @init4tech/ryans )
  - After a successful Flashbots bundle submission (that meets the deadline), the block transaction is forwarded to Pylon via pylon.post_blob_tx() to preload the sidecar.
  - Moves build_bundle_body from FlashbotsTask to SimResult for better encapsulation.
  - Adds submission deadline tracking — counters and logging distinguish between on-time vs late submissions, with early returns on failures or missed deadlines.
  - Flattens the submission async block — replaces nested match arms with early-return guards for readability.
  - Dependency updates — upgrades reth-chainspec to v1.11.0, renames TxCacheBundle → CachedBundle, fixes CacheResponse API change, and converts std::time::Instant →
  tokio::time::Instant.

### DIFF
--- a/src/tasks/submit/flashbots.rs
+++ b/src/tasks/submit/flashbots.rs
@@ -210,7 +210,7 @@ impl FlashbotsTask {
                     if Instant::now() > deadline {
                         counter!("signet.builder.flashbots.deadline_missed").increment(1);
                         warn!(
-                            hash = resp.as_ref().map(|r| r.bundle_hash.to_string()),
+                            ?resp,
                             "Submitted MEV bundle to Flashbots AFTER deadline - submission may be too late"
                         );
                         return;


src/tasks/submit/flashbots.rs
```diff
@@ -49,62 +54,36 @@ impl FlashbotsTask {
...
-        self.track_outbound_tx(&block_tx);
-
-        // Encode the transaction
-        let tx_bytes = block_tx.encoded_2718().into();
-
-        // Build the bundle body with the block_tx bytes as the last transaction in the bundle.
+    /// Builds a MEV bundle from a signed transaction envelope and simulation result.
+    fn build_bundle(&self, envelope: &TxEnvelope, sim_result: &SimResult) -> EthSendBundle {
+        let tx_bytes: Bytes = envelope.encoded_2718().into();
         let txs = self.build_bundle_body(sim_result, tx_bytes);
```
prestwich: refactor while we're here: the `build_bundle_body` method doesn't use self. which implies that it is on the wrong type. what type should it be on?

PR #224, opened 2026-02-13
fix: auto-sanitize TX_POOL_URL trailing slash

init4samwise:
## Summary

Automatically ensures `TX_POOL_URL` has a trailing slash during config sanitization, fixing `url.join()` behavior for endpoint construction.

## Problem

The Rust `url.join()` method behaves differently depending on whether the base URL ends with a trailing slash:
- `http://example.com/api/` + `join("transactions")` = `http://example.com/api/transactions` ✅
- `http://example.com/api` + `join("transactions")` = `http://example.com/transactions` ❌

This caused a confusing requirement in the README for users to manually add trailing slashes.

## Solution

- Add `sanitize()` method to `BuilderConfig` that ensures trailing slash
- Apply sanitization on config load in `config_from_env()`
- Remove manual trailing slash requirement from README
- Add 6 unit tests for URL sanitization

## Testing

Added unit tests covering:
- URLs without trailing slash get sanitized
- URLs with trailing slash remain unchanged
- URLs with paths (with/without trailing slash)
- Verify `url.join()` works correctly after sanitization

Closes [ENG-1531](https://linear.app/initiates/issue/ENG-1531)

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -304,8 +304,6 @@ impl BuilderConfig {
 
 #[cfg(test)]
 mod tests {
-    use super::*;
-
     /// Tests that URL sanitization correctly handles trailing slashes for url.join() behavior.
     ///
     /// The `url.join()` method behaves differently based on trailing slashes:


prestwich:
@init4samwise CI is broken

PR #233, opened 2026-03-02
fix: run block simulation on dedicated tokio runtime

rswanson:
## Summary

- The simulator was running CPU-heavy EVM simulation (`BlockBuild::build()`) on the main tokio runtime, which also serves the `/healthcheck` endpoint. With `CONCURRENCY_LIMIT=10` and a 2-CPU pod, all tokio worker threads were saturated with EVM work, preventing the healthcheck from responding within the 1s liveness probe timeout — causing repeated SIGKILL (exit code 137) and a crash loop.
- This creates a dedicated `tokio::runtime::Runtime` for simulation, sized to `CONCURRENCY_LIMIT` worker threads. The main runtime stays free for healthchecks, cache I/O, and channel operations regardless of simulation load.
- No throughput loss — simulation still runs with full concurrency on its own thread pool.

## Test plan

- [ ] Verify `make clippy` and `make test` pass (confirmed locally)
- [ ] Deploy to parmigiana and confirm the pod no longer crash-loops
- [ ] Verify liveness/readiness probes pass consistently during simulation
- [ ] Confirm block building completes within the 12s slot window

### DIFF
--- a/src/tasks/block/sim.rs
+++ b/src/tasks/block/sim.rs
@@ -14,6 +14,7 @@ use signet_sim::{BlockBuild, BuiltBlock, SimCache};
 use signet_types::constants::SignetSystemConstants;
 use std::time::{Duration, Instant};
 use tokio::{
+    runtime::Runtime,
     sync::{
         mpsc::{self},
         watch,
@@ -138,6 +139,7 @@ impl SimulatorTask {
         sim_items: SimCache,
         finish_by: Instant,
         sim_env: &SimEnv,
+        sim_rt: &Runtime,
     ) -> eyre::Result<BuiltBlock> {
         let concurrency_limit = self.config.concurrency_limit();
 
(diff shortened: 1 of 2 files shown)


Evalir (changes requested):
So, this technically _would_ fix the main issue: The main tokio thread's getting blocked by the simulation work, which _is_ blocking _and_ CPU heavy. The problem is that we're spawning an entirely new tokio runtime for this which increases resource usage just for isolating the sim task. On top of this, assigning `CONCURRENCY_LIMIT` worker threads is a tad wasteful since the sim runtime only runs one task at a time, which is the simulation work.

What we actually want is to isolate the single task, CPU heavy work from blocking the main tokio thread so that it can schedule handling healthchecks and other tasks while also continuing the sim work. Tokio can do this with `spawn_blocking`, which shifts the spawned task into its blocking task thread pool, avoiding the contention issue. I've implemented the changes on a PR on top of this: see #234

Evalir:
* **#234** <a href="https://app.graphite.com/github/pr/init4tech/builder/234?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#233** <a href="https://app.graphite.com/github/pr/init4tech/builder/233?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/init4tech/builder/233?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Evalir:
yep

PR #235, opened 2026-03-03
feat(bundle): filter bundles with stale host tx nonces before SimCache

init4samwise:
## Summary

Adds nonce checking for host transactions in BundlePoller, following the existing TxPoller pattern.

## Problem

BundlePoller ingests bundles without checking if host tx nonces are stale. Bundles with consumed nonces:
1. Get added to SimCache
2. Fail simulation with "nonce too low"
3. Get removed
4. Re-ingested on next poll (~1s)

This creates ERROR log spam and wastes simulation cycles.

## Solution

Before adding bundles to the outbound channel, spawn an async task that:
1. Decodes each host transaction
2. Recovers the sender address
3. Fetches current nonce from host provider
4. Drops bundle if any host tx has stale nonce (tx_nonce < provider_nonce)

This mirrors the existing `spawn_check_nonce` pattern in TxPoller.

## Changes

- Added `spawn_check_bundle_nonces` method to BundlePoller
- Added `decode_tx` helper function
- Updated task loop to use nonce checking before forwarding bundles
- Uses DEBUG-level logging for dropped bundles (not ERROR)

## Testing

CI will run cargo check/clippy. Manual testing can verify bundles with stale nonces are filtered.

---

Closes ENG-1937

### DIFF
--- a/src/tasks/cache/bundle.rs
+++ b/src/tasks/cache/bundle.rs
@@ -1,5 +1,12 @@
 //! Bundler service responsible for fetching bundles and sending them to the simulator.
 use crate::config::BuilderConfig;
+use alloy::{
+    consensus::{Transaction, transaction::SignerRecoverable},
+    eips::Decodable2718,
+    primitives::Bytes,
+    providers::Provider,
+    rlp::Buf,
+};
 use futures_util::{TryFutureExt, TryStreamExt};
 use init4_bin_base::perms::tx_cache::{BuilderTxCache, BuilderTxCacheError};
 use signet_tx_cache::{TxCacheError, types::CachedBundle};
@@ -8,7 +15,7 @@ use tokio::{
     task::JoinHandle,
     time::{self, Duration},
 };
-use tracing::{Instrument, debug, trace, trace_span, warn};
+use tracing::{Instrument, debug, debug_span, trace, trace_span, warn};
 
 /// Poll interval for the bundle poller in milliseconds.
 const POLL_INTERVAL_MS: u64 = 1000;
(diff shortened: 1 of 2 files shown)


init4samwise:
**[Claude Code]**

Note: The ticket also mentions reducing the log level at [signet-sdk driver.rs#L250](https://github.com/init4tech/signet-sdk/blob/26f660fb7c45033c471691734e18a30d10ae02ff/crates/bundle/src/send/driver.rs#L250) from ERROR to DEBUG. That change is in a different repo (signet-sdk) and would need a separate PR. This PR addresses the root cause by filtering stale bundles before they reach SimCache, which should eliminate the log spam at its source.

prestwich (changes requested):

src/tasks/cache/bundle.rs
```diff
@@ -71,6 +78,84 @@ impl BundlePoller {
...
+
+            let Ok(host_provider) =
+                crate::config().connect_host_provider().instrument(span.clone()).await
+            else {
+                span_debug!(span, "Failed to connect to host provider, stopping nonce check task");
+                return;
+            };
+
+            // Check each host transaction's nonce
+            for (idx, host_tx_bytes) in bundle.bundle.host_txs.iter().enumerate() {
```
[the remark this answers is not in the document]
init4samwise: Done in commits 6747eaf and 945b009. The implementation now uses `try_join_all` to fetch on-chain nonces for all unique signers concurrently. This achieves the same concurrent execution as FuturesUnordered while being more ergonomic for our use case (collecting results then validating sequentially).

src/tasks/cache/bundle.rs
```diff
@@ -94,10 +142,7 @@ impl BundlePoller {
                 histogram!("signet.builder.cache.bundles_fetched").record(bundles.len() as f64);
                 trace!(count = bundles.len(), "fetched bundles from tx-cache");
                 for bundle in bundles {
-                    if let Err(err) = outbound.send(bundle) {
-                        debug!(?err, "Failed to send bundle - channel is dropped");
-                        break;
-                    }
+                    Self::spawn_check_bundle_nonces(bundle, outbound.clone());
```
prestwich, line 152: minor TOCTOU on the outbound. it may have closed after the await

this is not a big deal, but may cause unnecessary resource usage validating nonces after the receiver is gone

src/tasks/cache/bundle.rs
```diff
@@ -59,6 +60,53 @@ impl BundlePoller {
...
+        tokio::spawn(async move {
+            let recovered = match bundle.bundle.try_to_recovered() {
+                Ok(recovered) => recovered,
+                Err(error) => {
+                    span_debug!(span, ?error, "Failed to recover bundle, dropping");
+                    return;
+                }
+            };
+
+            if recovered.host_txs().is_empty() {
```
prestwich, line 76: nested ifs. should be functional if possible

src/tasks/cache/bundle.rs
```diff
@@ -59,6 +60,53 @@ impl BundlePoller {
...
+                crate::config().connect_host_provider().instrument(span.clone()).await
+            else {
+                span_debug!(span, "Failed to connect to host provider, dropping bundle");
+                return;
+            };
+
+            let source = ProviderStateSource(host_provider);
+            match check_bundle_tx_list(recovered.host_tx_reqs(), &source).await {
+                Ok(SimItemValidity::Now) | Ok(SimItemValidity::Future) => {
+                    if outbound.send(bundle).is_err() {
```
prestwich: nit; functional

src/tasks/cache/bundle.rs
```diff
@@ -59,6 +60,53 @@ impl BundlePoller {
...
+
+            let Ok(host_provider) =
+                crate::config().connect_host_provider().instrument(span.clone()).await
+            else {
+                span_debug!(span, "Failed to connect to host provider, dropping bundle");
+                return;
+            };
+
+            let source = ProviderStateSource(host_provider);
+            match check_bundle_tx_list(recovered.host_tx_reqs(), &source).await {
```
prestwich: this future should be instrumented, right?

prestwich (changes requested):
some minor things

src/tasks/cache/bundle.rs
Fraser999: Good point!  Addressed in [c72ca42](https://github.com/init4tech/builder/pull/235/commits/c72ca4208a42b56819064ce5d3b7009798ff8915); there's still a toctou issue, but the window is mostly closed now.

src/tasks/cache/bundle.rs
Fraser999: Done in [c72ca42](https://github.com/init4tech/builder/pull/235/commits/c72ca4208a42b56819064ce5d3b7009798ff8915).

src/tasks/cache/bundle.rs
Fraser999: Done in [c72ca42](https://github.com/init4tech/builder/pull/235/commits/c72ca4208a42b56819064ce5d3b7009798ff8915).

src/tasks/cache/bundle.rs
Fraser999: Done in [c72ca42](https://github.com/init4tech/builder/pull/235/commits/c72ca4208a42b56819064ce5d3b7009798ff8915).

prestwich (approved):

PR #259, opened 2026-04-02
feat: replace TxPoller polling with SSE streaming

Evalir:
## Summary

Replaces the 1s timer-based polling loop in `TxPoller` with SSE streaming for real-time transaction delivery from the tx-pool. The new task lifecycle:

1. **Startup**: full paginated fetch of all transactions currently in the cache
2. **Steady state**: SSE stream (`/transactions/feed`) pushes new transactions as they arrive — no more redundant refetches
3. **Block env change**: full refetch to ensure consistency (covers any items the SSE stream may have missed)

On SSE disconnect or error, the poller reconnects with exponential backoff (1s initial, doubling up to 30s cap) and does a full refetch to cover the gap. Backoff resets on each successfully received transaction.

## Changes

- **`Cargo.toml`**: enable `sse` feature on `init4-bin-base` (transitively enables `signet-tx-cache/sse`)
- **`src/tasks/cache/tx.rs`**: rewrite `TxPoller` — replace poll loop with `full_fetch()` + `subscribe()` + `select!` over SSE items and block env changes. Add `reconnect()` with exponential backoff. Remove `poll_interval_ms`, `poll_duration()`, `Default` impl.
- **`src/tasks/cache/system.rs`**: pass `block_env` watch receiver to `TxPoller::new()`
- **`tests/tx_poller_test.rs`**: update integration test to use `TxCache` directly (no more `check_tx_cache()` method)

`BundlePoller` is unchanged — the `/bundles/feed` server endpoint is not yet available.

## Test plan

- [x] `make clippy` passes clean
- [x] `make test` — all 8 unit tests pass, integration tests correctly ignored
- [ ] Manual test against tx-pool: verify SSE subscription log, real-time tx delivery, refetch on block env change, reconnect with backoff on disconnect

### DIFF
--- a/src/tasks/cache/tx.rs
+++ b/src/tasks/cache/tx.rs
@@ -85,29 +85,35 @@ impl TxPoller {
         });
     }
 
-    /// Fetches all transactions from the cache, forwarding each to nonce
-    /// checking before it reaches the cache task.
-    async fn full_fetch(&self, outbound: &mpsc::UnboundedSender<ReceivedTx>) {
-        let span = trace_span!("TxPoller::full_fetch", url = %self.config.tx_pool_url);
+    /// Pulls every transaction currently in the cache, paginating until the
+    /// stream is exhausted. Pure fetch — no metrics, no dispatch.
+    async fn check_tx_cache(&self) -> Result<Vec<TxEnvelope>, TxCacheError> {
+        self.tx_cache.stream_transactions().try_collect().await
+    }
+
+    /// Fetches all transactions from the cache and dispatches each one to
+    /// a nonce-check task. Records poll metrics around the fetch.
+    async fn fetch_and_dispatch(&self, outbound: &mpsc::UnboundedSender<ReceivedTx>) {
+        let span = trace_span!("TxPoller::fetch_and_dispatch", url = %self.config.tx_pool_url);
 
         crate::metrics::inc_tx_poll_count();
-        if let Ok(transactions) = self
-            .tx_cache
-            .stream_transactions()
-            .try_collect::<Vec<_>>()
+        let Ok(transactions) = self
+            .check_tx_cache()
             .inspect_err(|error| {
                 crate::metrics::inc_tx_poll_errors();
                 debug!(%error, "Error fetching transactions");
             })
             .instrument(span.clone())
             .await
-        {
-            let _guard = span.entered();
-            crate::metrics::record_txs_fetched(transactions.len());
-            trace!(count = transactions.len(), "found transactions");
-            for tx in transactions {
-                self.spawn_check_nonce(tx, outbound.clone());
-            }
+        else {
+            return;
+        };
+
+        let _guard = span.entered();
+        crate::metrics::record_txs_fetched(transactions.len());
+        trace!(count = transactions.len(), "found transactions");
+        for tx in transactions {
+            self.spawn_check_nonce(tx, outbound.clone());
         }
     }
 
@@ -141,7 +147,7 @@ impl TxPoller {
             _ = time::sleep(*backoff) => {}
         }
         *backoff = (*backoff * 2).min(MAX_RECONNECT_BACKOFF);
-        let (_, stream) = tokio::join!(self.full_fetch(outbound), self.subscribe());
+        let (_, stream) = tokio::join!(self.fetch_and_dispatch(outbound), self.subscribe());
         stream
     }
 
@@ -179,7 +185,7 @@ impl TxPoller {
 
     async fn task_future(mut self, outbound: mpsc::UnboundedSender<ReceivedTx>) {
         // Initial full fetch of all transactions currently in the cache.
-        self.full_fetch(&outbound).await;
+        self.fetch_and_dispatch(&outbound).await;
 
         // Open the SSE stream for real-time delivery of new transactions.
         let mut sse_stream = self.subscribe().await;
@@ -202,7 +208,7 @@ impl TxPoller {
                         break;
                     }
                     trace!("Block env changed, refetching all transactions");
-                    self.full_fetch(&outbound).await;
+                    self.fetch_and_dispatch(&outbound).await;
                 }
             }
         }
--- a/src/tasks/cache/tx.rs
+++ b/src/tasks/cache/tx.rs
@@ -5,9 +5,8 @@ use alloy::{
     providers::Provider,
 };
 use futures_util::{Stream, StreamExt, TryFutureExt, TryStreamExt};
-use init4_bin_base::deps::metrics::{counter, histogram};
 use signet_tx_cache::{TxCache, TxCacheError};
-use std::{pin::Pin, time::Duration};
+use std::{ops::ControlFlow, pin::Pin, time::Duration};
 use tokio::{
     sync::{mpsc, watch},
     task::JoinHandle,
@@ -16,6 +15,9 @@ use tokio::{
 use tracing::{Instrument, debug, debug_span, trace, trace_span, warn};
 
 type SseStream = Pin<Box<dyn Stream<Item = Result<TxEnvelope, TxCacheError>> + Send>>;
+
+const INITIAL_RECONNECT_BACKOFF: Duration = Duration::from_secs(1);
+const MAX_RECONNECT_BACKOFF: Duration = Duration::from_secs(30);
 
 /// Implements a poller for the block builder to pull transactions from the
 /// transaction pool.
@@ -33,9 +35,6 @@ pub struct TxPoller {
 /// and on each block environment change, and subscribes to an SSE stream
 /// for real-time delivery of new transactions in between.
 impl TxPoller {
-    const INITIAL_RECONNECT_BACKOFF: Duration = Duration::from_secs(1);
-    const MAX_RECONNECT_BACKOFF: Duration = Duration::from_secs(30);
-
     /// Returns a new [`TxPoller`] with the given block environment receiver.
     pub fn new(envs: watch::Receiver<Option<SimEnv>>) -> Self {
         let config = crate::config();
@@ -91,20 +90,20 @@ impl TxPoller {
     async fn full_fetch(&self, outbound: &mpsc::UnboundedSender<ReceivedTx>) {
         let span = trace_span!("TxPoller::full_fetch", url = %self.config.tx_pool_url);
 
-        counter!("signet.builder.cache.tx_poll_count").increment(1);
+        crate::metrics::inc_tx_poll_count();
         if let Ok(transactions) = self
             .tx_cache
             .stream_transactions()
             .try_collect::<Vec<_>>()
             .inspect_err(|error| {
-                counter!("signet.builder.cache.tx_poll_errors").increment(1);
+                crate::metrics::inc_tx_poll_errors();
                 debug!(%error, "Error fetching transactions");
             })
             .instrument(span.clone())
             .await
         {
             let _guard = span.entered();
-            histogram!("signet.builder.cache.txs_fetched").record(transactions.len() as f64);
+            crate::metrics::record_txs_fetched(transactions.len());
             trace!(count = transactions.len(), "found transactions");
             for tx in transactions {
                 self.spawn_check_nonce(tx, outbound.clone());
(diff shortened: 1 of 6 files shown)


Evalir:
* **#273** <a href="https://app.graphite.com/github/pr/init4tech/builder/273?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#269** <a href="https://app.graphite.com/github/pr/init4tech/builder/269?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#259** <a href="https://app.graphite.com/github/pr/init4tech/builder/259?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/init4tech/builder/259?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/tasks/cache/tx.rs
```diff
@@ -96,46 +86,111 @@ impl TxPoller {
...
+        self.full_fetch(&outbound).await;
+
+        // Open the SSE stream for real-time delivery of new transactions.
+        let mut sse_stream = self.subscribe().await;
+        let mut backoff = Self::INITIAL_RECONNECT_BACKOFF;
+
+        loop {
+            tokio::select! {
+                item = sse_stream.next() => {
+                    match item {
```
prestwich: too nested

src/tasks/cache/tx.rs
```diff
@@ -96,46 +86,111 @@ impl TxPoller {
...
-                    self.spawn_check_nonce(tx, outbound.clone());
+    /// Reconnects the SSE stream with backoff. Performs a full refetch to
+    /// cover any items missed while disconnected.
+    async fn reconnect(
+        &mut self,
+        outbound: &mpsc::UnboundedSender<ReceivedTx>,
+        backoff: &mut Duration,
+    ) -> SseStream {
+        tokio::select! {
+            _ = time::sleep(*backoff) => {}
```
prestwich: this should be biased with an explanation of the choice of bias

src/tasks/cache/tx.rs
```diff
@@ -96,46 +86,111 @@ impl TxPoller {
...
-            // Check this here to avoid making the web request if we know
-            // we don't need the results.
-            if outbound.is_closed() {
-                span.in_scope(|| trace!("No receivers left, shutting down"));
-                break;
+    /// Opens an SSE subscription to the transaction feed. Returns an empty
+    /// stream on connection failure so the caller can handle reconnection
+    /// uniformly.
+    async fn subscribe(&self) -> SseStream {
+        match self.tx_cache.subscribe_transactions().await {
```
prestwich: rewrite functionally

src/tasks/cache/tx.rs
```diff
@@ -96,46 +86,111 @@ impl TxPoller {
...
+        backoff: &mut Duration,
+    ) -> SseStream {
+        tokio::select! {
+            _ = time::sleep(*backoff) => {}
+            // Break the sleep early on block env change or channel close —
+            // full_fetch below serves the same purpose the env arm would have.
+            _ = self.envs.changed() => {}
+        }
+        *backoff = (*backoff * 2).min(Self::MAX_RECONNECT_BACKOFF);
+        self.full_fetch(outbound).await;
```
prestwich: these can be joined, no?

src/tasks/cache/tx.rs
```diff
@@ -96,46 +86,111 @@ impl TxPoller {
...
+        outbound: &mpsc::UnboundedSender<ReceivedTx>,
+        backoff: &mut Duration,
+    ) -> SseStream {
+        tokio::select! {
+            _ = time::sleep(*backoff) => {}
+            // Break the sleep early on block env change or channel close —
+            // full_fetch below serves the same purpose the env arm would have.
+            _ = self.envs.changed() => {}
+        }
+        *backoff = (*backoff * 2).min(Self::MAX_RECONNECT_BACKOFF);
```
prestwich: i don't love putting the exponential backoff in-line instead of using an existing implementation, or having it be an unbounded number of attempts. at what point is a failure deemed permanent?

src/tasks/cache/tx.rs
```diff
@@ -1,60 +1,50 @@
...
+/// [`TxPoller`] fetches transactions from the transaction pool on startup
+/// and on each block environment change, and subscribes to an SSE stream
+/// for real-time delivery of new transactions in between.
 impl TxPoller {
-    /// Returns a new [`TxPoller`] with the given config.
-    /// * Defaults to 1000ms poll interval (1s).
-    pub fn new() -> Self {
-        Self::new_with_poll_interval_ms(POLL_INTERVAL_MS)
-    }
+    const INITIAL_RECONNECT_BACKOFF: Duration = Duration::from_secs(1);
```
prestwich: top of file instead of assoc consts?

src/tasks/cache/tx.rs
```diff
@@ -96,46 +86,111 @@ impl TxPoller {
...
+    async fn full_fetch(&self, outbound: &mpsc::UnboundedSender<ReceivedTx>) {
+        let span = trace_span!("TxPoller::full_fetch", url = %self.config.tx_pool_url);
+
+        counter!("signet.builder.cache.tx_poll_count").increment(1);
+        if let Ok(transactions) = self
+            .tx_cache
+            .stream_transactions()
+            .try_collect::<Vec<_>>()
+            .inspect_err(|error| {
+                counter!("signet.builder.cache.tx_poll_errors").increment(1);
```
prestwich: in #263 we reified metrics under new patterns. these should match

src/tasks/cache/tx.rs
```diff
@@ -96,46 +86,111 @@ impl TxPoller {
...
-        self.tx_cache.stream_transactions().try_collect().await
+    /// Fetches all transactions from the cache, forwarding each to nonce
+    /// checking before it reaches the cache task.
+    async fn full_fetch(&self, outbound: &mpsc::UnboundedSender<ReceivedTx>) {
+        let span = trace_span!("TxPoller::full_fetch", url = %self.config.tx_pool_url);
+
+        counter!("signet.builder.cache.tx_poll_count").increment(1);
+        if let Ok(transactions) = self
+            .tx_cache
+            .stream_transactions()
```
prestwich: sdk API thing. we now have "stream transactions" and "subscribe", which are not clear about their behavior

src/tasks/cache/tx.rs
```diff
@@ -96,46 +86,111 @@ impl TxPoller {
         });
     }
 
-    /// Polls the transaction cache for transactions, paginating through all available pages.
-    pub async fn check_tx_cache(&self) -> Result<Vec<TxEnvelope>, TxCacheError> {
-        self.tx_cache.stream_transactions().try_collect().await
+    /// Fetches all transactions from the cache, forwarding each to nonce
+    /// checking before it reaches the cache task.
+    async fn full_fetch(&self, outbound: &mpsc::UnboundedSender<ReceivedTx>) {
```
prestwich: architectural:
why was `check_tx_cache` deleted if its logic is repeated inline here?

This function also does more than fetch, it dispatches tasks. So its name should reflect that

prestwich (changes requested):

src/tasks/cache/tx.rs
Evalir: filthy claude

src/tasks/cache/tx.rs
Evalir: Yeah I don't love this personally. How about `fetch_` vs `watch_`?

src/tasks/cache/tx.rs
Evalir: there wasn't a principled decision why here. the fn just got absorbed into the refactored big fetch function. Aside from what claude thinks, I think it's better to keep it separate just for the sake of keeping everything else maintainable.

src/tasks/cache/tx.rs
```diff
@@ -95,46 +85,138 @@ impl TxPoller {
...
+        ControlFlow::Continue(())
+    }
+
+    async fn task_future(mut self, outbound: mpsc::UnboundedSender<ReceivedTx>) {
+        // Initial full fetch of all transactions currently in the cache.
+        self.fetch_and_dispatch(&outbound).await;
 
-            time::sleep(self.poll_duration()).await;
+        // Open the SSE stream for real-time delivery of new transactions.
+        let mut sse_stream = self.subscribe().await;
```
Fraser999: Shouldn't this be more like the `reconnect` impl where we run both futures in a `join!` like
```
let (_, mut sse_stream) = tokio::join!(self.fetch_and_dispatch(outbound), self.subscribe());
```

src/tasks/cache/tx.rs
```diff
@@ -1,59 +1,49 @@
...
-    }
+    /// Receiver for block environment updates, used to trigger refetches.
+    envs: watch::Receiver<Option<SimEnv>>,
 }
 
-/// [`TxPoller`] implements a poller task that fetches transactions from the transaction pool
-/// and sends them into the provided channel sender.
+/// [`TxPoller`] fetches transactions from the transaction pool on startup
+/// and on each block environment change, and subscribes to an SSE stream
+/// for real-time delivery of new transactions in between.
```
Fraser999: We could maybe move this to replace the now slightly out-of-date doc comment on the struct definition (it mentions "poller").

src/tasks/cache/tx.rs
```diff
@@ -95,46 +85,138 @@ impl TxPoller {
...
+                    {
+                        break;
+                    }
+                }
+                res = self.envs.changed() => {
+                    if res.is_err() {
+                        debug!("Block env channel closed, shutting down");
+                        break;
+                    }
+                    trace!("Block env changed, refetching all transactions");
```
Fraser999: nit: I think this maybe warrants being at debug level

Fraser999 (commented):
We could add a counter to metrics for the number of SSE stream reconnect attempts.

src/tasks/cache/tx.rs
Evalir: yeah, do agree

Fraser999 (approved):
This LGTM now, with the proviso that I think [this comment](https://github.com/init4tech/builder/pull/259#discussion_r3088472424) potentially isn't resolved.

src/tasks/cache/tx.rs
Evalir: So, for failures being deemed permanent, I think keeping this unbounded might make sense: there's still a periodic full tx cache fetch in place, so allowing the SSE stream to just try reconnecting until it succeeds would allow the task to self-heal without needing to restart the builder.

Didn't notice backon was a dep—we're using that now

src/tasks/cache/tx.rs
```diff
@@ -1,59 +1,61 @@
...
 
-/// Poll interval for the transaction poller in milliseconds.
-const POLL_INTERVAL_MS: u64 = 1000;
+const INITIAL_RECONNECT_BACKOFF: Duration = Duration::from_secs(1);
+const MAX_RECONNECT_BACKOFF: Duration = Duration::from_secs(30);
+
+fn reconnect_backoff() -> ExponentialBackoff {
+    ExponentialBuilder::default()
+        .with_factor(2.0)
+        .with_min_delay(INITIAL_RECONNECT_BACKOFF)
```
prestwich, line 33: so looking at this, the sequence would be `1,2,4,8,16,32,32,...` and never stop, right? the behavior should be documented. Is 1 second a good starting place? or should it be shorter?

src/tasks/cache/tx.rs
```diff
@@ -95,46 +97,132 @@ impl TxPoller {
...
+    /// should shut down.
+    async fn handle_sse_item(
+        &mut self,
+        item: Option<Result<TxEnvelope, TxCacheError>>,
+        outbound: &mpsc::UnboundedSender<ReceivedTx>,
+        stream: &mut SseStream,
+    ) -> ControlFlow<()> {
+        match item {
+            Some(Ok(tx)) => {
+                self.reconnect_backoff = reconnect_backoff();
```
prestwich: nit: this should probably go below the shutdown check

also, I don't love that `backon` doesn't expose a `.reset()` on their backoff and forces you to reconstruct. there's a small space/time optimization to make by caching the builder and using `builder.build()` repeatedly instead of re-making the builder each time. but i think it doesn't matter at all

src/tasks/cache/tx.rs
```diff
@@ -95,46 +97,132 @@ impl TxPoller {
...
-            time::sleep(self.poll_duration()).await;
+    async fn task_future(mut self, outbound: mpsc::UnboundedSender<ReceivedTx>) {
+        // Initial full fetch of all currently-cached transactions, plus SSE
+        // subscription for real-time delivery.
+        let (_, mut sse_stream) =
+            tokio::join!(self.fetch_and_dispatch(&outbound), self.subscribe());
+
+        loop {
+            tokio::select! {
+                item = sse_stream.next() => {
```
prestwich, line 209: i think this arm can be collapsed a bit by adding stream combinators that transform the `sse_stream` into a future that only ever returns on `ControlFlow::Break`? 

sketch:

```
let stream = sse_stream
   .then(|item| self.handle_sse_item(...))
   .skip_while(|flow| flow.is_continue)
```

prestwich: passing the `&mut sse_stream` as an arg break this tho

prestwich: but that's a bit of a code smell anyway, so it might be worth re-flowing to remove it by adding more information in the ControlFLow

prestwich (approved):
some architecture discussion comments. no code blockers, so approving

PR #269, opened 2026-04-23
feat: replace BundlePoller polling with SSE streaming

Evalir:
## Description

Mirrors #259 (tx-poller SSE) for bundles. Replaces the 1s polling loop in `BundlePoller` with an SSE subscription to `/bundles/feed` via `BuilderTxCache::subscribe_bundles` (newly exposed by the SDK, matched by the tx-pool-webservice endpoint).

The structure matches `TxPoller` exactly:
- Initial `full_fetch` paginates through `stream_bundles` to seed the cache on startup.
- `subscribe` opens the SSE stream and yields `CachedBundle`s in real time.
- On stream error / stream end, reconnects with exponential backoff (1s → 30s cap), racing the sleep against the block-env watcher so an env change wins and triggers a fresh full refetch immediately.
- On block-env change, the full_fetch runs again to cover anything missed during disconnection.
- `NotOurSlot` is logged at trace level (expected when the builder is not slot-permissioned) in both the full-fetch and subscribe paths — avoids spurious warn-level noise.

The public `check_bundle_cache()` wrapper is dropped; the integration test now uses `BuilderTxCache::stream_bundles` directly, matching the style of `tx_poller_test.rs`.

## Related Issue

Stacked on #259.

## Testing

- [x] `make fmt` passes
- [x] `make clippy` passes
- [x] `make test` passes
- [x] `cargo doc --no-deps` passes with `-D warnings`
- [x] `cargo test --features test-utils --no-run` builds all integration tests
- [ ] Exercise against a live tx-pool-webservice to verify SSE reconnect and env-driven refetch behavior end-to-end

### DIFF
--- a/src/tasks/cache/bundle.rs
+++ b/src/tasks/cache/bundle.rs
@@ -143,13 +143,13 @@ impl BundlePoller {
             .inspect(
                 |_| debug!(url = %self.config.tx_pool_url, "SSE bundle subscription established"),
             )
-            .inspect_err(|error| {
-                crate::metrics::inc_sse_subscribe_errors();
-                match error {
-                    BuilderTxCacheError::TxCache(TxCacheError::NotOurSlot) => {
-                        trace!("Not our slot to subscribe to bundles");
-                    }
-                    _ => warn!(%error, "Failed to open SSE bundle subscription"),
+            .inspect_err(|error| match error {
+                BuilderTxCacheError::TxCache(TxCacheError::NotOurSlot) => {
+                    trace!("Not our slot to subscribe to bundles");
+                }
+                _ => {
+                    crate::metrics::inc_sse_subscribe_errors();
+                    warn!(%error, "Failed to open SSE bundle subscription");
                 }
             })
             .ok()
@@ -188,6 +188,23 @@ impl BundlePoller {
         }
     }
 
+    /// Reconnects and swaps in the fresh stream, or returns `Break` if the
+    /// outbound channel closed during the reconnect loop.
+    async fn try_reconnect(
+        &mut self,
+        outbound: &mpsc::UnboundedSender<CachedBundle>,
+        backoff: &mut Duration,
+        stream: &mut SseStream,
+    ) -> ControlFlow<()> {
+        match self.reconnect(outbound, backoff).await {
+            Some(s) => {
+                *stream = s;
+                ControlFlow::Continue(())
+            }
+            None => ControlFlow::Break(()),
+        }
+    }
+
     /// Returns `Break` when the outbound channel has closed and the task
     /// should shut down.
     async fn handle_sse_item(
@@ -205,23 +222,17 @@ impl BundlePoller {
                     return ControlFlow::Break(());
                 }
                 Self::spawn_check_bundle_nonces(bundle, outbound.clone());
+                ControlFlow::Continue(())
             }
             Some(Err(error)) => {
                 warn!(%error, "SSE bundle stream interrupted, reconnecting");
-                let Some(s) = self.reconnect(outbound, backoff).await else {
-                    return ControlFlow::Break(());
-                };
-                *stream = s;
+                self.try_reconnect(outbound, backoff, stream).await
             }
             None => {
                 warn!("SSE bundle stream ended, reconnecting");
-                let Some(s) = self.reconnect(outbound, backoff).await else {
-                    return ControlFlow::Break(());
-                };
-                *stream = s;
+                self.try_reconnect(outbound, backoff, stream).await
             }
         }
-        ControlFlow::Continue(())
     }
 
     async fn task_future(mut self, outbound: mpsc::UnboundedSender<CachedBundle>) {
--- a/src/tasks/cache/bundle.rs
+++ b/src/tasks/cache/bundle.rs
@@ -1,60 +1,72 @@
 //! Bundler service responsible for fetching bundles and sending them to the simulator.
-use crate::config::BuilderConfig;
-use futures_util::{TryFutureExt, TryStreamExt};
+use crate::{config::BuilderConfig, tasks::env::SimEnv};
+use futures_util::{Stream, StreamExt, TryFutureExt, TryStreamExt};
 use init4_bin_base::perms::tx_cache::{BuilderTxCache, BuilderTxCacheError};
 use signet_sim::{ProviderStateSource, SimItemValidity, check_bundle_tx_list};
 use signet_tx_cache::{TxCacheError, types::CachedBundle};
+use std::{ops::ControlFlow, pin::Pin, time::Duration};
 use tokio::{
-    sync::mpsc::{UnboundedReceiver, UnboundedSender, unbounded_channel},
+    sync::{mpsc, watch},
     task::JoinHandle,
-    time::{self, Duration},
+    time,
 };
-use tracing::{Instrument, debug_span, trace, trace_span, warn};
+use tracing::{Instrument, debug, debug_span, trace, trace_span, warn};
 
-/// Poll interval for the bundle poller in milliseconds.
-const POLL_INTERVAL_MS: u64 = 1000;
+type SseStream = Pin<Box<dyn Stream<Item = Result<CachedBundle, BuilderTxCacheError>> + Send>>;
 
-/// The BundlePoller polls the tx-pool for bundles.
+const INITIAL_RECONNECT_BACKOFF: Duration = Duration::from_secs(1);
+const MAX_RECONNECT_BACKOFF: Duration = Duration::from_secs(30);
+
+/// The BundlePoller fetches bundles from the tx-pool on startup and on each
+/// block environment change, and subscribes to an SSE stream for real-time
+/// delivery of new bundles in between.
 #[derive(Debug)]
 pub struct BundlePoller {
     /// The builder configuration values.
     config: &'static BuilderConfig,
-
     /// Client for the tx cache.
     tx_cache: BuilderTxCache,
-
-    /// Defines the interval at which the bundler polls the tx-pool for bundles.
-    poll_interval_ms: u64,
+    /// Receiver for block environment updates, used to trigger refetches.
+    envs: watch::Receiver<Option<SimEnv>>,
 }
 
-impl Default for BundlePoller {
-    fn default() -> Self {
-        Self::new()
-    }
-}
-
-/// Implements a poller for the block builder to pull bundles from the tx-pool.
 impl BundlePoller {
-    /// Creates a new BundlePoller from the provided builder config.
-    pub fn new() -> Self {
-        Self::new_with_poll_interval_ms(POLL_INTERVAL_MS)
-    }
-
-    /// Creates a new BundlePoller from the provided builder config and with the specified poll interval in ms.
-    pub fn new_with_poll_interval_ms(poll_interval_ms: u64) -> Self {
+    /// Returns a new [`BundlePoller`] with the given block environment receiver.
+    pub fn new(envs: watch::Receiver<Option<SimEnv>>) -> Self {
         let config = crate::config();
         let tx_cache = BuilderTxCache::new(config.tx_pool_url.clone(), config.oauth_token());
-        Self { config, tx_cache, poll_interval_ms }
+        Self { config, tx_cache, envs }
     }
 
-    /// Returns the poll duration as a [`Duration`].
-    const fn poll_duration(&self) -> Duration {
-        Duration::from_millis(self.poll_interval_ms)
-    }
+    async fn full_fetch(&self, outbound: &mpsc::UnboundedSender<CachedBundle>) {
+        let span = trace_span!("BundlePoller::full_fetch", url = %self.config.tx_pool_url);
 
-    /// Fetches all bundles from the tx-cache, paginating through all available pages.
-    pub async fn check_bundle_cache(&self) -> Result<Vec<CachedBundle>, BuilderTxCacheError> {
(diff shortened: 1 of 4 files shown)


Evalir:
* **#273** <a href="https://app.graphite.com/github/pr/init4tech/builder/273?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#269** <a href="https://app.graphite.com/github/pr/init4tech/builder/269?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/init4tech/builder/269?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#259** <a href="https://app.graphite.com/github/pr/init4tech/builder/259?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/tasks/cache/bundle.rs
```diff
@@ -1,113 +1,184 @@
...
-                        warn!(%error, "Failed to fetch bundles from tx-cache");
-                    }
-                })
-                .instrument(span.clone())
-                .await
-            else {
-                time::sleep(self.poll_duration()).await;
-                continue;
-            };
+    async fn task_future(mut self, outbound: mpsc::UnboundedSender<CachedBundle>) {
```
dylanlott, line 238: should span this and pass the span to the lifecycle methods as necessary

src/tasks/cache/bundle.rs
```diff
@@ -1,113 +1,184 @@
...
+                }
+                _ => warn!(%error, "Failed to open SSE bundle subscription"),
+            })
+            .map(|s| Box::pin(s) as SseStream)
+            .unwrap_or_else(|_| Box::pin(futures_util::stream::empty()))
+    }
 
-            // Check this here to avoid making the web request if we know
-            // we don't need the results.
-            if outbound.is_closed() {
```
dylanlott, line 123: This check is removed from this arm, but it's still called in the tx cache arm. Need to check `outbound.is_closed` somewhere in this arm. In its current state, the simulator could drop its receiver while the SSE is in the error or none branch and would repeat its reconnection loop forever.

src/tasks/cache/bundle.rs
dylanlott: and as I think about it more, we probably want to add at least the block number to span's log fields.

src/tasks/cache/bundle.rs
```diff
@@ -1,113 +1,184 @@
...
+            .subscribe_bundles()
+            .await
+            .inspect(
+                |_| debug!(url = %self.config.tx_pool_url, "SSE bundle subscription established"),
+            )
+            .inspect_err(|error| match error {
+                BuilderTxCacheError::TxCache(TxCacheError::NotOurSlot) => {
+                    trace!("Not our slot to subscribe to bundles");
+                }
+                _ => warn!(%error, "Failed to open SSE bundle subscription"),
```
dylanlott: should add a metric for SSE failures in the subscribe loop to match the `inc_tx_poll_errors()` in the fetch and dispatch loop.

src/tasks/cache/bundle.rs
```diff
@@ -1,113 +1,184 @@
...
-                span.in_scope(|| trace!("No receivers left, shutting down"));
-                break;
+    /// Runs a full refetch concurrently with re-subscribing, to cover any
+    /// items missed while disconnected.
+    async fn reconnect(
+        &mut self,
+        outbound: &mpsc::UnboundedSender<CachedBundle>,
+        backoff: &mut Duration,
+    ) -> SseStream {
+        tokio::select! {
```
dylanlott: should add a metric like `inc_tx_sse_reconnects()` as well

dylanlott (changes requested):
We should test these changes in Parmigiana asap before taking to mainnet.

src/tasks/cache/bundle.rs
```diff
@@ -1,113 +1,184 @@
...
+                    {
+                        break;
+                    }
+                }
+                res = self.envs.changed() => {
+                    if res.is_err() {
+                        debug!("Block env channel closed, shutting down");
                         break;
                     }
+                    trace!("Block env changed, refetching all bundles");
```
Fraser999: Maybe worth changing to `debug!`?  I guess be consistent with how the corresponding log line is handled for the tx stream (ref [this comment](https://github.com/init4tech/builder/pull/259#discussion_r3154159954)).

src/tasks/cache/bundle.rs
```diff
@@ -1,113 +1,184 @@
...
+                    trace!("No receivers left, shutting down");
+                    return ControlFlow::Break(());
+                }
+            }
+            Some(Err(error)) => {
+                warn!(%error, "SSE bundle stream interrupted, reconnecting");
+                *stream = self.reconnect(outbound, backoff).await;
             }
+            None => {
+                warn!("SSE bundle stream ended, reconnecting");
```
Fraser999, line 232: **[Claude Code]**

When `subscribe()` (line 98) fails to open the connection, it logs `"Failed to open SSE bundle subscription"` and returns `stream::empty()` as a sentinel. The first `sse_stream.next()` then yields `None` immediately, which lands here and emits `"SSE bundle stream ended, reconnecting"`. Two warns per failed subscribe cycle, and the second one is a lie - the stream never opened.

Suggest having `subscribe()` return `Option<SseStream>` (`.ok().map(...)` instead of `unwrap_or_else(empty)`) and adding a `None`-branch in `task_future`/`reconnect` that just schedules a retry without going through `handle_sse_item`. That keeps the "stream ended" warn for the case it actually describes (a real stream that closed mid-flight) and avoids double-logging.

src/tasks/cache/bundle.rs
```diff
@@ -1,113 +1,184 @@
...
 
-    /// Fetches all bundles from the tx-cache, paginating through all available pages.
-    pub async fn check_bundle_cache(&self) -> Result<Vec<CachedBundle>, BuilderTxCacheError> {
-        self.tx_cache.stream_bundles().try_collect().await
+    /// Fetches all bundles from the cache and forwards each to the outbound
+    /// channel. Records poll metrics around the fetch.
+    async fn fetch_and_forward(&self, outbound: &mpsc::UnboundedSender<CachedBundle>) {
+        let span = trace_span!("BundlePoller::fetch_and_forward", url = %self.config.tx_pool_url);
+
+        crate::metrics::inc_bundle_poll_count();
```
Fraser999, line 50: This metric (and other ones with "poll" in their names) should maybe be renamed as they don't really measure individual poll attempts any more.

Fraser999 (commented):
We could use actual integration tests for this stream and the tx stream (I think the existing one is more like a connectivity smoke test).  Not blocking, but worth a follow-up ticket?

src/tasks/cache/bundle.rs
Evalir: **[Claude Code]**

Done in 13b25a4 — `task_future` now carries an `#[instrument]` span with `url` and `block_number` fields. `block_number` is recorded on startup and re-recorded on every env-change branch so the field tracks the current block. The lifecycle methods (`fetch_and_forward`, `subscribe`, `reconnect`, `handle_sse_item`) inherit this span automatically via async context propagation, so I dropped the redundant per-method `trace_span!`s.

src/tasks/cache/bundle.rs
Evalir: **[Claude Code]**

Done in 13b25a4 — added an `if outbound.is_closed() { break; }` guard at the top of `task_future`'s select loop, plus an early `outbound.is_closed()` check at the top of `reconnect`'s loop body. Together those break the forever-reconnect cycle: if the simulator drops its receiver while we're stuck in a stream error/None loop, the next iteration of either loop exits cleanly.

src/tasks/cache/bundle.rs
Evalir: **[Claude Code]**

Done in 13b25a4 — added `inc_sse_subscribe_errors` and called it from `subscribe`'s `inspect_err`. Used a generic name (no `bundle_` prefix) to match the existing `inc_sse_reconnect_attempts` pattern; tx.rs can adopt the same helper later without renaming.

src/tasks/cache/bundle.rs
Evalir: **[Claude Code]**

Done in 13b25a4 — `BundlePoller::reconnect` now increments the existing `inc_sse_reconnect_attempts` counter once per attempt, matching how `TxPoller::reconnect` uses it. Also generalized the metric's description from "SSE transaction stream reconnect attempts" to "SSE stream reconnect attempts" since both pollers share it.

src/tasks/cache/bundle.rs
Evalir: **[Claude Code]**

Done in 13b25a4 — bumped the env-change refetch log from `trace!` to `debug!`, matching the agreed change on tx.rs.

src/tasks/cache/bundle.rs
Evalir: **[Claude Code]**

Done in 13b25a4 — `subscribe()` now returns `Option<SseStream>`; on `None`, the caller (initial path in `task_future` or the Err/None arms of `handle_sse_item`) hands off to `reconnect`, which loops internally with backoff until either a real stream is established (`Some`) or the outbound channel closes (`None`, signalling the task to shut down). The misleading "stream ended, reconnecting" warn no longer fires after a failed subscribe — the only warn comes from `subscribe`'s actual error path.

src/tasks/cache/bundle.rs
Evalir: **[Claude Code]**

Acknowledged. Deferring this rename to its own PR — it cuts across `BundlePoller` and `TxPoller` metrics plus any dashboards/alerts, so I'd rather not mix it with the SSE work here. Will track as a follow-up.

src/tasks/cache/bundle.rs
```diff
@@ -1,113 +1,225 @@
...
+                *stream = s;
+            }
+        }
+        ControlFlow::Continue(())
+    }
+
+    #[instrument(
+        skip_all,
+        fields(url = %self.config.tx_pool_url, block_number = tracing::field::Empty),
+    )]
```
Fraser999: This is a long-running task under normal conditions, so shouldn't be instrumented like this I believe.  Maybe something more like how we do instrumentation in `SubmitTask::task_future` would work better?  That would also avoid the call to `Span::current()` inside `record_block_number` which can be flaky depending on the runtime log level chosen.

src/tasks/cache/bundle.rs
```diff
@@ -1,113 +1,225 @@
...
-            time::sleep(self.poll_duration()).await;
         }
     }
 
-    /// Spawns a task that sends bundles it finds to its channel sender.
-    pub fn spawn(self) -> (UnboundedReceiver<CachedBundle>, JoinHandle<()>) {
-        let (outbound, inbound) = unbounded_channel();
+    fn record_block_number(&self) {
+        if let Some(env) = self.envs.borrow().as_ref() {
+            Span::current().record("block_number", env.rollup_block_number());
```
Fraser999: I think we use `sim.ru.number` for this elsewhere rather than `block_number`.

src/tasks/cache/bundle.rs
Evalir: **[Claude Code]**

Done in 53e49ff — dropped the `#[instrument]` from `task_future` and the `record_block_number` helper that called `Span::current().record(...)`. The env-change branch now captures `env.clone_span()` (the `BlockConstruction` span built by `EnvTask`) and `.instrument()`s the inline async block that runs the refetch — same shape as `SubmitTask::task_future`'s per-work-item span and `CacheTask::task_future`'s `env.span().enter()`. No more long-lived task-wide span and no more `Span::current()` lookup.

src/tasks/cache/bundle.rs
Evalir: **[Claude Code]**

Resolved in 53e49ff — by reusing the `BlockConstruction` span via `env.clone_span()` instead of recording a custom field, the existing `sim.ru.number` / `sim.host.number` / `sim.slot` fields on that span attach to refetch logs automatically. The standalone `record_block_number` helper (and its `block_number` field) is gone.

src/tasks/cache/bundle.rs
```diff
@@ -1,113 +1,224 @@
...
+    /// would double-log "stream ended" on a failure that never opened.
+    async fn subscribe(&self) -> Option<SseStream> {
+        self.tx_cache
+            .subscribe_bundles()
+            .await
+            .inspect(
+                |_| debug!(url = %self.config.tx_pool_url, "SSE bundle subscription established"),
+            )
+            .inspect_err(|error| {
+                crate::metrics::inc_sse_subscribe_errors();
```
dylanlott: This will fire on a `NotOurSlot` error. Move this down to the match below so we don't conflate `NotOurSlot` errors.

src/tasks/cache/bundle.rs
```diff
@@ -1,113 +1,224 @@
...
+                *stream = s;
+            }
+            None => {
+                warn!("SSE bundle stream ended, reconnecting");
+                let Some(s) = self.reconnect(outbound, backoff).await else {
+                    return ControlFlow::Break(());
+                };
+                *stream = s;
+            }
+        }
```
dylanlott, line 235: Nit: The `None` and `Some(Err(_))` branches are nearly identical - could make a `try_reconnect` method and condense these down to just log and invoke.

Fraser999 (approved):

dylanlott (approved):

PR #280, opened 2026-06-09
fix(cache): rank ingested items with the live env basefee, not 0

Evalir:
## Summary

`CacheTask::task_future` declared `basefee` as a loop-local that was reset to `0` at the top of every `select!` iteration and assigned (`sim_env.basefee`) only in the `envs.changed()` arm. Because `add_bundle`/`add_tx` run in the *other* `select!` arms — i.e. separate loop iterations — they always received `basefee = 0`. Cached items were therefore ranked by `effective_gas_price(0)` instead of the env basefee.

## Impact

The basefee feeds only `SimItem::calculate_total_fee`, whose result is the cache's **ranking key** — a heuristic for simulation/inclusion order and the eviction key when the cache is over capacity. It does **not** affect block validity or execution (the real basefee is applied by the EVM during building).

So the effect is limited to ordering/eviction quality:
- **Negligible on gouda** (basefee ~7 wei) — the ranking is effectively identical to correct, which is why this is *not* the bundle-inclusion failure.
- **Latent** on any higher-basefee chain or congested period, where the missing `basefee · gas` term can reorder gas-heavy items or evict the wrong item under capacity pressure → at worst a marginally less profitable block.

## Fix

Read the block number and basefee from the **live env** at ingest time via a new `current_block_and_basefee()` helper, used by both ingest arms; the loop-local `basefee = 0` is removed.

Closes ENG-2333

## Test

`cache_task_ranks_with_env_basefee` drives a real `CacheTask`, publishes an env with a known basefee, ingests a tx whose effective tip depends on the basefee, and asserts the cached rank matches `calculate_total_fee(env_basefee)` (with a guard that the basefee actually changes the rank, so it fails on the bug).

## Validation

`cargo build`, the new test, `cargo +nightly fmt`, `cargo clippy --lib --all-targets` (no warnings), and the cache-module tests all pass.


Evalir:
* **#280** <a href="https://app.graphite.com/github/pr/init4tech/builder/280?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/init4tech/builder/280?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/tasks/cache/task.rs
```diff
@@ -186,3 +204,75 @@ impl IngestionSummary {
...
+
+        let (_env_tx, env_rx) = watch::channel(Some(sim_env));
+        let (_bundle_tx, bundle_rx) = mpsc::unbounded_channel();
+        let (txn_tx, txn_rx) = mpsc::unbounded_channel();
+
+        let (cache, _jh) = CacheTask::new(env_rx, bundle_rx, txn_rx).spawn();
+
+        // A tx whose effective tip depends on the basefee: with max_fee 100 gwei
+        // and tip 50 gwei, the tip is min(50, 100-60)=40 gwei at BASEFEE vs
+        // 50 gwei at basefee 0, so the two ranks differ.
```
Fraser999:
```suggestion
        // A tx whose rank depends on the basefee: rank is effective_gas_price
        // (min(max_fee, basefee + tip)) * gas_limit, i.e. min(100, 60+50)=100 gwei
        // per gas at BASEFEE vs min(100, 0+50)=50 gwei at basefee 0.
```

Fraser999 (approved):

